PR #1, opened 2022-10-21
Setup initial cross compilation workflow

ianks:
This is a PR to setup cross gem compilation. Right now, it just adds the ability to cross compile gems via Github UI. We'll figure out publishing later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Remove Gemfile.lock
         run: rm Gemfile.lock
 
-      - uses: oxidize-rb/actions/setup-ruby-and-rust@v1
+      - uses: oxidize-rb/actions/setup-ruby-and-rust@main
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
@@ -34,3 +34,12 @@ jobs:
 
       - name: Lint rust
         run: cargo clippy && cargo fmt --check
+
+      - name: Smoke test gem install
+        shell: bash
+        run: |
+          bundle exec rake build
+          gem install pkg/wasmtime-rb-*.gem --verbose
+          script="puts Wasmtime::Engine.new(Wasmtime::Config.new).precompile_module('(module)')"
+          ruby -rwasmtime -e "$script" | grep wasmtime-aot
+          echo "✅ Successfully gem installed"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,12 +34,3 @@ jobs:
 
       - name: Lint rust
         run: cargo clippy && cargo fmt --check
-
-      - name: Smoke test gem install
-        shell: bash
-        run: |
-          bundle exec rake build
-          gem install pkg/wasmtime-rb-*.gem --verbose
-          script="puts Wasmtime::Engine.new(Wasmtime::Config.new).precompile_module('(module)')"
-          ruby -rwasmtime -e "$script" | grep wasmtime-aot
-          echo "✅ Successfully gem installed"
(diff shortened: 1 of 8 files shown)


ianks:
[cha-ching!](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3301371024)

.github/workflows/ci.yml
```diff
@@ -34,3 +34,12 @@ jobs:
 
       - name: Lint rust
         run: cargo clippy && cargo fmt --check
+
+      - name: Smoke test gem install
+        shell: bash
+        run: |
+          bundle exec rake build
+          gem install pkg/wasmtime-rb-*.gem --verbose
+          script="puts Wasmtime::Engine.new(Wasmtime::Config.new).precompile_module('(module)')"
```
jbourassa, line 43: Nit: the config is now, this should also work:
```suggestion
          script="puts Wasmtime::Engine.new.precompile_module('(module)')"
```

jbourassa (commented):
This is great, thank you so much Ian.

I see the `cross-compile`'s smoke test fails:

> ```
> ERROR:  Error installing pkg/wasmtime-rb-0.1.0-x86_64-linux.gem:
>	wasmtime-rb-0.1.0-x86_64-linux requires Ruby version >= 3.0, < > 3.1.dev. The current ruby version is 3.1.2.20.
>```

Looks like we're only building for 3.0.0:

> no configuration section for specified version of Ruby (rbconfig-x86_64-linux-"3.1.0)
>no configuration section for specified version of Ruby (rbconfig-x86_64-linux-2.7.0")

Searching for that error message [led me to `rake-compiler`](https://github.com/rake-compiler/rake-compiler/blob/8c907ac32f87bf16da6e84105fb4d6bf257e6bbd/lib/rake/extensiontask.rb#L408-L411). Looks like `config.yml` doesn't have entries for all rubies. Is it because `xrubies` bundles 1 ruby version per image, unlike `rake-compiler-docker`?

ianks:
@jbourassa I am very confused about this as well. Not sure where the bug is… Gonna contact rake-compiler team

.github/workflows/cross-gem.yml
```diff
@@ -0,0 +1,45 @@
...
+          cargo-cache: true
+          cargo-vendor: true
+          cache-version: v1
+
+      - uses: oxidize-rb/cross-gem-action@v6
+        with:
+          platform: ${{ matrix.ruby-platform }}
+          version: "0.9.34"
+          env:
+            RUBY_CC_VERSION="3.1.0:3.0.0:2.7.0"
```
flavorjones, line 37: This should be

```
          env:
            RUBY_CC_VERSION: "3.1.0:3.0.0:2.7.0"
```

flavorjones:
The current error, I think, is because rake-compiler will put the cross-compiled shared object file into a subdirectory named after the ruby minor version (e.g., `lib/wasmtime/3.1/ext.so` instead of `lib/wasmtime/ext.so`).

You can see this in the build log: https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3340442663/jobs/5530540697#step:4:280

You should look first for a ruby-specific precompiled .so, and fall back to `lib/wasmtime/ext.so` if you can't find it (which means the user compiled upon installation). A lot of this is explained in https://github.com/flavorjones/ruby-c-extensions-explained/tree/main/precompiled

An example of how to do this is

https://github.com/flavorjones/ruby-c-extensions-explained/tree/main/precompiled

also 

https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/extension.rb

Hope this helps!

jbourassa (approved):

PR #2, opened 2022-10-27
Implement `Caller#export`

jbourassa:
Implement `Caller#export` which allows getting a instance's memories or funcs from a `Caller`.

This is a a little odd because the caller itself only "lives" for the duration of call, but we can't control the lifetime of Ruby objects we hand out:

```ruby
Func.new(...) do |caller|
  $yes_i_can_do_this = caller
end
```

The approach I'm taking here is to model a Caller's "context" as an `Option`. While the func is being called, it'll be `Some`, but we'll change it to `None` before exiting the func call.

Thus extracting the context from a `Caller` can fail. That's implemented with an `Error` that gets surfaced as  Ruby exception. See the added test.

---

This PR also makes it easy to implement funcrefs: https://github.com/bytecodealliance/wasmtime-rb/compare/caller-extern...funcref-bindings

### DIFF
--- a/ext/src/ruby_api/func.rs
+++ b/ext/src/ruby_api/func.rs
@@ -1,9 +1,10 @@
 use super::{
-    convert::{ToRubyValue, ToWasmVal},
+    convert::{ToRubyValue, ToWasmVal, WrapWasmtimeType},
+    externals::Extern,
     func_type::FuncType,
     params::Params,
     root,
-    store::{Store, StoreData},
+    store::{Store, StoreContextValue, StoreData},
 };
 use crate::{error, helpers::WrappedStruct};
 use magnus::{
@@ -11,22 +12,39 @@ use magnus::{
     value::BoxValue, DataTypeFunctions, Error, Exception, Module as _, Object, RArray, RClass,
     RString, TryConvert, TypedData, Value, QNIL,
 };
-use std::cell::RefCell;
+use std::cell::UnsafeCell;
 use wasmtime::{
-    AsContextMut, Caller as CallerImpl, Extern, ExternType, Func as FuncImpl, Trap, Val,
+    AsContext, AsContextMut, Caller as CallerImpl, Func as FuncImpl, StoreContext, StoreContextMut,
+    Trap, Val,
 };
 
 /// @yard
+/// @rename Wasmtime::Func
 /// Represents a WebAssembly Function
 /// @see https://docs.rs/wasmtime/latest/wasmtime/struct.Func.html Wasmtime's Rust doc
-#[derive(TypedData, Debug)]
-#[magnus(class = "Wasmtime::Func", mark, size, free_immediatly)]
-pub struct Func {
-    store: WrappedStruct<Store>,
+#[derive(Debug)]
+pub struct Func<'a> {
+    store: StoreContextValue<'a>,
     inner: FuncImpl,
 }
 
-impl DataTypeFunctions for Func {
+unsafe impl<'a> TypedData for Func<'a> {
+    fn class() -> magnus::RClass {
+        *memoize!(RClass: root().define_class("Func", Default::default()).unwrap())
+    }
+
+    fn data_type() -> &'static magnus::DataType {
+        memoize!(magnus::DataType: {
+            let mut builder = DataTypeBuilder::<Func<'_>>::new("Wasmtime::Func");
+            builder.size();
+            builder.mark();
+            builder.free_immediatly();
+            builder.build()
+        })
+    }
+}
+
+impl DataTypeFunctions for Func<'_> {
     fn mark(&self) {
         self.store.mark()
     }
@@ -40,9 +58,9 @@ struct ShareableProc(Proc);
 unsafe impl Send for ShareableProc {}
 unsafe impl Sync for ShareableProc {}
 
-unsafe impl Send for Func {}
+unsafe impl Send for Func<'_> {}
 
-impl Func {
+impl<'a> Func<'a> {
     /// @yard
     /// @def new(store, type, callable, &block)
     /// @param store [Store]
@@ -80,12 +98,12 @@ impl Func {
         let inner = wasmtime::Func::new(context, ty.clone(), make_func_closure(ty, callable));
 
         Ok(Self {
-            store: wrapped_store,
+            store: wrapped_store.into(),
             inner,
         })
     }
 
-    pub fn from_inner(store: WrappedStruct<Store>, inner: FuncImpl) -> Self {
+    pub fn from_inner(store: StoreContextValue<'a>, inner: FuncImpl) -> Self {
         Self { store, inner }
     }
 
@@ -107,24 +125,24 @@ impl Func {
     ///   * 1 => +Object+
     ///   * > 1 => +Array<Object>+
     pub fn call(&self, args: &[Value]) -> Result<Value, Error> {
-        let store: &Store = self.store.try_convert()?;
-        Self::invoke(store, &self.inner, args).map_err(|e| e.into())
+        Self::invoke(&self.store, &self.inner, args).map_err(|e| e.into())
     }
 
     pub fn invoke(
-        store: &Store,
+        store: &StoreContextValue,
         func: &wasmtime::Func,
         args: &[Value],
     ) -> Result<Value, InvokeError> {
-        let func_ty = func.ty(store.context_mut());
+        let func_ty = func.ty(store.context_mut()?);
         let param_types = func_ty.params().collect::<Vec<_>>();
         let params = Params::new(args, param_types)?.to_vec()?;
         let mut results = vec![Val::null(); func_ty.results().len()];
 
-        func.call(store.context_mut(), &params, &mut results)
+        func.call(store.context_mut()?, &params, &mut results)
             .map_err(|e| {
                 store
                     .context_mut()
+                    .expect("store context is still reachable")
                     .data_mut()
                     .take_last_error()
                     .unwrap_or_else(|| error!("Could not invoke function: {}", e))
@@ -144,7 +162,7 @@ impl Func {
     }
 }
 
-impl From<&Func> for Extern {
+impl From<&Func<'_>> for wasmtime::Extern {
     fn from(func: &Func) -> Self {
         Self::Func(func.get())
     }
@@ -158,25 +176,27 @@ pub fn make_func_closure(
     let ty = ty.to_owned();
     let callable = ShareableProc(callable);
 
-    move |caller: CallerImpl<'_, StoreData>, params: &[Val], results: &mut [Val]| {
-        let caller = RefCell::new(caller);
+    move |caller_impl: CallerImpl<'_, StoreData>, params: &[Val], results: &mut [Val]| {
+        let caller_value: Value = Caller::new(caller_impl).into();
+        let caller = caller_value.try_convert::<&Caller>().unwrap();
 
         let rparams = RArray::with_capacity(params.len() + 1);
-        rparams.push(Caller { inner: &caller }).ok();
+        rparams.push(caller_value).unwrap();
 
         for (i, param) in params.iter().enumerate() {
             let rparam = param.to_ruby_value().map_err(|e| {
                 wasmtime::Trap::new(format!("invalid argument at index {}: {}", i, e))
             })?;
-            rparams.push(rparam).ok();
+            rparams.push(rparam).unwrap();
         }
 
         let callable = callable.0;
-        callable
+
+        let result = callable
             .call(unsafe { rparams.as_slice() })
             .map_err(|e| {
                 if let Error::Exception(exception) = e {
-                    caller.borrow_mut().data_mut().exception().hold(exception);
+                    caller.hold_exception(exception);
                 }
                 e
             })
@@ -210,7 +230,13 @@ pub fn make_func_closure(
                     callable.inspect(),
                     e
                 ))
-            })
+            });
+
+        // Drop the wasmtime::Caller so it does not outlive the Func call, if e.g. the user
+        // assigned the Ruby Wasmtime::Caller instance to a global.
+        caller.expire();
+
+        result
     }
 }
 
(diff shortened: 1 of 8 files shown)


spec/unit/func_spec.rb
```diff
@@ -113,9 +154,5 @@ def build_func(params, results, &block)
       store = Store.new(engine, {})
       Func.new(store, FuncType.new(params, results), &block)
     end
-
-    # Used to test that you can send a `Method` object with `method(:foo)`
-    def noop
-    end
   end
```
jbourassa, line 157: Not relevant to this PR but I came across this dead code.

ext/src/ruby_api/store.rs
```diff
@@ -95,6 +95,51 @@ impl Store {
...
+                // The Caller is on the stack while it's "live". Right before the end of a host call,
+                // we remove the Caller form the Ruby object, thus there is no need to mark.
+            }
+        }
+    }
+
+    pub fn context(&self) -> Result<StoreContext<StoreData>, Error> {
+        match self {
+            Self::Store(s) => {
+                let store = s.try_convert::<&Store>().expect("a Store typed data");
```
ianks, line 119: Would it be better to:

```suggestion
                let store = s.try_convert::<&Store>()?;
```

ext/src/ruby_api/func.rs
```diff
@@ -198,36 +212,67 @@ impl From<BoxValue<Exception>> for InvokeError {
     }
 }
 
-struct Caller<'a> {
-    inner: &'a RefCell<CallerImpl<'a, StoreData>>,
+pub struct Caller<'a> {
+    inner: UnsafeCell<Option<CallerImpl<'a, StoreData>>>,
```
ianks, line 216: Probably worth a comment as to why we use `UnsafeCell` here.

> We use `UnsafeCell` because we need interior mutability, but can't use dynamically checked `Refcell` because (... reasons? I assume GC?).

ianks: After thinking about it more, it might be nice to wrap this functionality in a new type. That way, we can specify the behavior a bit more if we need to. At the very least, it can help to communicate the purpose a bit more clearly.

```rust
/// A handle to a [`Caller`] which allows for multiple re-entant borrows (i.e. Ruby -> Wasm -> Ruby -> Wasm).
pub struct CallerHandle<'a> {
  value: UnsafeCell<Option<CallerImpl<'a, StoreData>>>
}

impl <'a>CallerHandle<'a> {
  pub fn get_mut(&self) -> Somethin {
    // Add logic here to keep track of borrowers if possible...

    unsafe { &mut *self.inner.get() }
      .as_mut()
      .ok_or_else(|| error!("Caller outlived its Func execution"))?
  }

  pub fn expire(&self) {
    unsafe { *self.value.get() = None }
  }
}

File: ext/src/ruby_api/func.rs [hunk 4]
```diff
@@ -198,36 +212,67 @@ impl From<BoxValue<Exception>> for InvokeError {
...
+            .map(|c| c.as_context())
+    }
+
+    pub fn context_mut(&self) -> Result<StoreContextMut<StoreData>, Error> {
+        unsafe { (*self.inner.get()).as_mut() }
+            .ok_or_else(|| error!("Caller outlived its Func execution"))
+            .map(|c| c.as_context_mut())
+    }
+
+    pub fn drop_inner(&self) {
```
saulecabrera, line 266: Do you think that there's value in renaming this function? Maybe to something like what @ianks suggests above, since we're not really dropping the value (in Rust's terms)

ext/src/ruby_api/func.rs
```diff
@@ -116,11 +121,12 @@ pub fn make_func_closure(
...
 
-    move |caller: CallerImpl<'_, StoreData>, params: &[Val], results: &mut [Val]| {
-        let caller = RefCell::new(caller);
+    move |caller_impl: CallerImpl<'_, StoreData>, params: &[Val], results: &mut [Val]| {
+        let caller_value: Value = Caller::new(caller_impl).into();
+        let caller = caller_value.try_convert::<&Caller>().unwrap();
 
         let rparams = RArray::with_capacity(params.len() + 1);
-        rparams.push(Caller { inner: &caller }).ok();
+        rparams.push(caller_value).ok();
```
saulecabrera, line 129: I'll note here but applies in some other places in the codebase. Why are we not expecting or unwrapping the result of the `push`?

ext/src/ruby_api/store.rs
```diff
@@ -95,6 +95,51 @@ impl Store {
     }
 }
 
+/// A wrapper around a Ruby Value that has a store context.
+/// Used in places where both Store or Caller can be used.
+#[derive(Debug)]
+pub enum StoreContextValue {
```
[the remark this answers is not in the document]
ianks: IMO, it's nicer to match against Rust enums than strings for a couple of reasons:

1. You get exhaustiveness checks
2. Avoid a call into libruby
3. You avoid an unnecessary string allocation (although probably trivial, it's nice to know your performance is bounded by a integer compare operation)

The potential downsides:

1. You are duplicating state instead of deriving in a sense, but in this case i'd probably prefer Rust being the source of truth. It takes some questions off the table that I'd rather not think about... (i.e what happens if a user inherits from `Wasmtime::Module`?)
2. You may need an extra `u32` to represent the state ([actually, this can likely be avoided if we use `RTypedData` instead of `Value` to make use of the non-zeroness](https://github.com/matsadler/magnus/blob/main/src/r_typed_data.rs#L46))

ext/src/ruby_api/func.rs
```diff
@@ -198,36 +212,67 @@ impl From<BoxValue<Exception>> for InvokeError {
...
+        Ok(Some(export))
+    }
+
+    pub fn context(&self) -> Result<StoreContext<StoreData>, Error> {
+        unsafe { (*self.inner.get()).as_ref() }
+            .ok_or_else(|| error!("Caller outlived its Func execution"))
+            .map(|c| c.as_context())
+    }
+
+    pub fn context_mut(&self) -> Result<StoreContextMut<StoreData>, Error> {
```
ianks, line 260: Could we implement [`AsContextMut`](https://docs.rs/wasmtime/latest/src/wasmtime/store/context.rs.html#118) instead?

jbourassa: That was my first attempt, but that wasn't possible because we can't get a `&mut self` from a Value, and that's what the trait requires:
>```rust
>fn as_context_mut(&mut self) -> StoreContextMut<'_, Self::Data>;
>```

Unless you can think of a way of working around that?this problem?

ext/src/ruby_api/func.rs
jbourassa: Oops yeah, makes sense -- `unwrap` is what we want.

The only way this can fail is if the `Array` is frozen, which can't really happen: we just created the Array above.

ext/src/ruby_api/store.rs
jbourassa: I initially went with `expect` to express that this is an invariant, regardless of what the Ruby program does.

The counter-argument is that we should avoid panic if we can 🤷. No strong opinion on my end, happy to defer to your Rust experience here.

ext/src/ruby_api/store.rs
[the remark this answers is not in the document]
saulecabrera: > against Rust enums than strings

Generally I'm questioning if the usage of an enum is what we want. I believe it would be nicer if we could avoid the introduction of this special, intermediate representation, and either: just use the `Value` representation or convert any special values (e.g. `Store`, `Caller`) into their Rust representations immediately after entering Rust and let the type system represent any invariants of what's possible with each (e.g. calling `context` `context_mut`), and just that consistently across modules/functions that require it.

ext/src/ruby_api/func.rs
jbourassa: I'm using `UnsafeCell` because I couldn't get it to compile with `RefCell`, not because of re-entrance. Callers don't need to be re-entrant because you get a new caller for each entrance, that shouldn't be a problem.

The problem I ran into is that I couldn't find a way to return a `wasmtime::store::context::StoreContext<_>` from a `RefCell`:


```rust
pub struct CallerRefCell<'a> {
    inner: RefCell<Option<wasmtime::Caller<'a, StoreData>>>,
}

impl<'a> CallerRefCell<'a> {
    pub fn context(&self) -> Result<StoreContext<StoreData>, Error> {
        match *self.inner.borrow() {
            Some(ref inner) => Ok(inner.as_context()),
            None => err!("Caller outlived its Func execution"),
        }
    }

    pub fn context_mut(&self) -> Result<StoreContextMut<StoreData>, Error> {
        match *self.inner.borrow_mut() {
            Some(ref mut inner) => Ok(inner.as_context_mut()),
            None => err!("Caller outlived its Func execution"),
        }
    }
}
```


```
error[E0515]: cannot return value referencing temporary value
   --> ext/src/ruby_api/func.rs:315:32
    |
314 |         match *self.inner.borrow() {
    |                ------------------- temporary value created here
315 |             Some(ref inner) => Ok(inner.as_context()),
    |                                ^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function

error[E0515]: cannot return value referencing temporary value
   --> ext/src/ruby_api/func.rs:322:36
    |
321 |         match *self.inner.borrow_mut() {
    |                ----------------------- temporary value created here
322 |             Some(ref mut inner) => Ok(inner.as_context_mut()),
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

Not knowing how to fix this I resorted to `UnsafeCell`. I have a hunch that `context` and `context_mut` are actually unsafe: the context can become invalid if the caller is expired (set the Option to None), and the type system wouldn't know about that. In practice it is probably not a problem, though, as the Wasmtime's context will still exist.

Let me know if you can think of a way to avoid resorting to `UnsafeCell`.

ext/src/ruby_api/store.rs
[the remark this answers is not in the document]
ianks: [And here is the solution I think I like](https://github.com/bytecodealliance/wasmtime-rb/commit/bdd8486d4f5439ff149366fc6f00e21592255338). It uses `RTypedData` instead of `Value` (to have a more specific constraint with `NonZeroValue`), and `TryFrom` implementation to avoid the enums everywhere. It still stores an enum, but it's always derived from the Ruby state.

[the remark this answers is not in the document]
jbourassa: > It still stores an enum, but it's always derived from the Ruby state.

One of the thing that was bothering Saúl was that it'd be "easy" to pass in incorrect value when creating the enum. This proposal partially addresses that, I like it.

---

We also explored using 2 other approaches:
1. Using a trait. We couldn't get it to work because there are places where we don't have a concrete type and can't introduce a generic (e.g. the `init` function). Maybe that would've been possible with a Box, but that added allocations and runtime dispatch, so we dropped the idea.
2. Using a Struct to store a `Value` and a enum of `&Caller` or `&Store`. I _think_ that would've worked, but require introducing lifetimes in _many_ places for little gain in the end.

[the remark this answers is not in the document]
ianks: > Using a Struct to store a Value and a enum of &Caller or &Store. I think that would've worked, but require introducing lifetimes in many places for little gain in the end.

That's essentially what `RTypedData` is. Only limitation is that it does not support generics (makes derivation much harder). Magnus also makes sure to call `rb_check_typeddata` when attempting to convert, so passing in incorrect data will error at runtime (which we need regardless of static type checks).

ext/src/ruby_api/func.rs
ianks: I cannot! Happy with the current approach.

ext/src/ruby_api/func.rs
jbourassa: I used your recommendation of wrapping the UnsafeCell into a new type, but nothing keeps track of borrows. I also cherry-picked your RTypedData commit (bdd8486d4f5439ff149366fc6f00e21592255338).

I (again) looked into using `RefCell` and failed. Maybe it'd work by _not_ implementing `context` / `context_mut` on `StoreContextValue`, and instead getting the context from the enum variants directly when needed (with a macro?). But that'd be clunky, 😞.

ext/src/ruby_api/store.rs
```diff
@@ -113,6 +115,52 @@ impl Store {
     }
 }
 
+/// A wrapper around a Ruby Value that has a store context.
+/// Used in places where both Store or Caller can be used.
+#[derive(Debug)]
+pub enum StoreContextValue<'a> {
+    Store(WrappedStruct<Store>),
+    Caller(WrappedStruct<Caller<'a>>),
+}
```
jbourassa, line 124: After rebasing on #14, I switched this to use `WrappedStruct`.

Pros:
- Cant' fail to convert, I went from `TryFrom<Value> for StoreContextValue` to `From<WrappedStruct<Store>> for StoreContextValue`.
- We only need to convert once.

Cons: `Func`, `Memory`, etc. all have to know about the lifetime. Which also means we can't derive `DataType` and have to use the more verbose `DataTypeBuilder`.

I think I like this better. But it's an easy thing to change, so 🤷.

ext/src/ruby_api/func.rs
jbourassa: I'd like to see how you did it if you still have that commit around.

saulecabrera (approved):
> ext/src/ruby_api/caller.rs

That file is empty; I believe we should remove it.

jbourassa:
> > ext/src/ruby_api/caller.rs
> 
> That file is empty; I believe we should remove it.

Oh yes! I created the file thinking I could extract `Caller` to it, but then changed my mind: in Wasmtime, Caller is in the same file as Func 🤷.

ext/src/ruby_api/func.rs
jbourassa: I'm going to merge this to keep moving forward, but happy to iterate on it once in main. Any and all feedback is still welcome 🙇.

PR #3, opened 2022-10-29
Add hard-mode CI step with `GC.stress`

ianks:
This PR adds a step in CI which runs tests with `GC.stress`. This should sus out any strange GC related bugs before they are merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,11 @@ jobs:
       - name: Run ruby tests
         run: bundle exec rake spec
 
+      - name: Run ruby tests (hard-mode with GC.stress)
+        run: bundle exec rake spec
+        env:
+          GC_STRESS: "true"
+
       - name: Lint ruby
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -29,6 +29,11 @@ jobs:
       - name: Run ruby tests
         run: bundle exec rake spec
 
+      - name: Run ruby tests (hard-mode with GC.stress)
+        run: bundle exec rake spec
+        env:
+          GC_STRESS: "true"
```
jbourassa, line 35: Good idea! 👍 

Is there value in doing this on all platform and ruby version? Because this is slow, I was thinking we could do it only once on e.g. linux + ruby 3.1.

ianks: It's not too slow, IMO. If it does get slow, we can change it to a limited matrix like you mentioned.

jbourassa (approved):

PR #4, opened 2022-10-31
Add `Module.deserialize_file`

ianks:
This PR adds support for [`wasmtime::Module::deserialize_file`](https://docs.wasmtime.dev/api/wasmtime/struct.Module.html#method.deserialize_file).

> This method is provided because it can be faster than [deserialize](https://docs.wasmtime.dev/api/wasmtime/struct.Module.html#method.deserialize) since the data doesn’t need to be copied around, but rather the module can be used directly from an mmap’d view of the file provided.

### DIFF
--- a/ext/src/ruby_api/module.rs
+++ b/ext/src/ruby_api/module.rs
@@ -4,7 +4,7 @@ use magnus::{function, method, Error, Module as _, Object, RString};
 use wasmtime::Module as ModuleImpl;
 
 #[derive(Clone)]
-#[magnus::wrap(class = "Wasmtime::Module")]
+#[magnus::wrap(class = "Wasmtime::Module", size, free_immediatly)]
 pub struct Module {
     inner: ModuleImpl,
 }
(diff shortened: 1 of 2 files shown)


ext/src/ruby_api/module.rs
```diff
@@ -4,7 +4,7 @@ use magnus::{function, method, Error, Module as _, Object, RString};
 use wasmtime::Module as ModuleImpl;
 
 #[derive(Clone)]
-#[magnus::wrap(class = "Wasmtime::Module")]
+#[magnus::wrap(class = "Wasmtime::Module", size, free_immediatly)]
```
ianks, line 7: Not related to the PR, but these methods can be safely derived. The typo is intentional (mis-spelled in magnus).

jbourassa (commented):
LGTM, outside of the CI failure on Windows:

```
  1) Wasmtime::Module.deserialize_file can deserialize a module from a file
     Failure/Error: mod = Module.deserialize_file(engine, tmpfile.path)

     Wasmtime::Error:
       Could not deserialize module from file: "failed to create file mapping for: D:/a/_temp/deserialize-file20221031-1376-ontron.so"
     # ./spec/unit/module_spec.rb:22:in `deserialize_file'
     # ./spec/unit/module_spec.rb:22:in `block (3 levels) in <module:Wasmtime>'
```

Wonder if that's because the file is still open? Trying something out in c6406dbed3e381d59c3057a002ca4f51649f9958, we'll see if that works.

ianks:
Looks like Windows works now, [but not on ruby-head](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3363799557/jobs/5577407913)....

PR #11, opened 2022-11-10
Add precompiled gems for `mingw`

ianks:
This PR adds precompiled gems for `mingw`, so Windows users do not have to install a Rust toolchain to use this gem.

I took the opportunity to refactor CI a bit to make sure we only support stable Ruby versions (2.7..3.1 as of today).

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -29,6 +29,8 @@ jobs:
       fail-fast: false
       matrix:
         ruby-platform: ${{ fromJSON(needs.ci-data.outputs.result).supported-ruby-platforms }}
+    env:
+      CROSS_PLATFORMS: ${{ matrix.ruby-platform }}
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -7,18 +7,28 @@ on:
     branches: ["main", "cross-gem/*"]
 
 jobs:
+  ci-data:
+    runs-on: ubuntu-latest
+    outputs:
+      result: ${{ steps.fetch.outputs.result }}
+    steps:
+      - id: fetch
+        uses: oxidize-rb/actions/fetch-ci-data@main
+        with:
+          supported-ruby-platforms: |
+            exclude:
+              - arm-linux # 32-bit not supported on magnus yet
+          stable-ruby-versions: |
+            exclude: [head]
+
   native:
     name: Build native gems
+    needs: ci-data
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        ruby-platform:
-          - "aarch64-linux"
-          - "arm64-darwin"
-          - "x86_64-darwin"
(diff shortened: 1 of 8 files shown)


ianks:
Just waiting on https://github.com/matsadler/magnus/pull/27 now

.github/workflows/build-gems.yml
```diff
@@ -7,18 +7,29 @@ on:
...
       matrix:
-        ruby-platform:
-          - "aarch64-linux"
-          - "arm64-darwin"
-          - "x86_64-darwin"
-          - "x86_64-linux"
-          - "x86_64-linux-musl"
+        ruby-platform: ${{ fromJSON(needs.ci-data.outputs.result).supported-ruby-platforms }}
+    env:
+      CROSS_PLATFORMS: ${{ matrix.ruby-platform }}
```
jbourassa, line 32: What read this? Couldn't find any reference to it in the gems (`rg CROSS_PLATFORMS $(bundle show --paths)`) nor the actions.

.github/workflows/build-gems.yml
```diff
@@ -7,18 +7,29 @@ on:
     branches: ["main", "cross-gem/*"]
 
 jobs:
+  ci-data:
+    runs-on: ubuntu-latest
+    outputs:
+      result: ${{ steps.fetch.outputs.result }}
+    steps:
+      - id: fetch
+        uses: oxidize-rb/actions/fetch-ci-data@main
```
jbourassa, line 16: 👌 really cool

rakelib/compile.rake
```diff
@@ -2,13 +2,7 @@ require "rake/extensiontask"
...
 GEMSPEC = Bundler.load_gemspec("wasmtime-rb.gemspec")
 
-CROSS_PLATFORMS = [
-  "aarch64-linux",
-  "arm64-darwin",
-  "x86_64-darwin",
-  "x86_64-linux",
-  "x86_64-linux-musl"
-]
+CROSS_PLATFORMS = [ENV["RUBY_TARGET"]].compact
```
jbourassa, line 5: If I get this right, it should've been that way from the beginning, right? We run the  compile task once per target as per the matrix, and we get the [`RUBY_TARGET` from the Dockerfiles](https://github.com/oxidize-rb/rb-sys/blob/fa0f82885d9fff747af024288636cb6e4eaf4eed/docker/setup/rubybashrc.sh#L54).

jbourassa (approved):
Thanks for this. Looks like the build is failing, not sure why, it works on my M1 mac but fails on CI for arm64-darwin.

Left some questions so that I can learn 🤓.

.github/workflows/build-gems.yml
ianks: Ahh, it's an old diff that is not needed anymore. Will delete.

rakelib/compile.rake
ianks: Yup exactly. Seems like the sane default for GitHub actions, would be nice to have that be the default for RCD somehow...

PR #13, opened 2022-11-11
YARD doc generation

jbourassa:
First pass at generating documentation using a [YARD plugin I wrote](https://github.com/oxidize-rb/yard-rustdoc). I manually generated the docs and put them on a GH pages for now: https://bytecodealliance.github.io/wasmtime-rb/v2.0.2-alpha1/

Decisions:
- Added an extra validation step to ensure everything that's documented actually exists. That way, a typo in class names or method names won't go unnoticed.
- Make any YARD warning an error, and build the docs in CI. The idea is that anything that breaks the doc shouldn't go in main.
- Document the Ruby specificities thoroughly, but document Wasmtime only briefly. The links to the official doc should suffice, no need to copy everything.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,8 @@ group :development do
   gem "standard", "~> 1.3"
   gem "get_process_mem"
   gem "ruby-lsp", require: false
+  gem "yard", require: false
+  gem "yard-rustdoc", "~> 0.1", require: false
 end
 
 group :test do
--- a/ext/src/ruby_api/func.rs
+++ b/ext/src/ruby_api/func.rs
@@ -9,13 +9,15 @@ use crate::error;
 use magnus::{
     block::Proc, function, gc, memoize, method, r_typed_data::DataTypeBuilder,
     scan_args::scan_args, value::BoxValue, DataTypeFunctions, Error, Exception, Module as _,
-    Object, RArray, RClass, RHash, RString, TryConvert, TypedData, Value, QNIL,
+    Object, RArray, RClass, RString, TryConvert, TypedData, Value, QNIL,
 };
 use std::cell::RefCell;
 use wasmtime::{
     AsContextMut, Caller as CallerImpl, Extern, ExternType, Func as FuncImpl, Trap, Val,
 };
 
+/// @yard
+/// Represents a WebAssembly Function
 #[derive(TypedData, Debug)]
 #[magnus(class = "Wasmtime::Func", mark, size, free_immediatly)]
 pub struct Func {
@@ -40,8 +42,18 @@ unsafe impl Sync for ShareableProc {}
 unsafe impl Send for Func {}
 
 impl Func {
+    /// @yard
+    ///
+    /// @def new(store, type, callable, &block)
+    /// @param store [Wasmtime::Store]
+    /// @param type [Wasmtime::FuncType]
+    /// @param block [Block] The funcs's implementation
+    /// @return [Wasmtime::Func]
+    ///
+    /// @example
+    ///   store = Wasmtime::Store.new(Wasmtime::Engine.new)
     pub fn new(args: &[Value]) -> Result<Self, Error> {
-        let args = scan_args::<(Value, &FuncType), (), (), (), RHash, Proc>(args)?;
+        let args = scan_args::<(Value, &FuncType), (), (), (), (), Proc>(args)?;
         let (s, functype) = args.required;
         let callable = args.block;
 
@@ -64,6 +76,8 @@ impl Func {
         self.inner
     }
 
+    /// @yard
+    /// Func#call method
     pub fn call(&self, args: &[Value]) -> Result<Value, Error> {
         let store: &Store = self.store.try_convert()?;
         Self::invoke(store, &self.inner, args).map_err(|e| e.into())
--- a/ext/src/ruby_api/func.rs
+++ b/ext/src/ruby_api/func.rs
@@ -18,6 +18,7 @@ use wasmtime::{
 
 /// @yard
 /// Represents a WebAssembly Function
+/// @see https://docs.rs/wasmtime/latest/wasmtime/struct.Func.html Wasmtime's Rust doc
 #[derive(TypedData, Debug)]
 #[magnus(class = "Wasmtime::Func", mark, size, free_immediatly)]
 pub struct Func {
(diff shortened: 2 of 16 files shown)


ext/src/ruby_api/func.rs
```diff
@@ -40,8 +43,27 @@ unsafe impl Sync for ShareableProc {}
...
+    ///   * 1 => +Object+
+    ///   * > 1 => +Array<Object>+
+    ///
+    /// @return [Func]
+    ///
+    /// @example
+    ///   store = Wasmtime::Store.new(Wasmtime::Engine.new)
     pub fn new(args: &[Value]) -> Result<Self, Error> {
-        let args = scan_args::<(Value, &FuncType), (), (), (), RHash, Proc>(args)?;
+        let args = scan_args::<(Value, &FuncType), (), (), (), (), Proc>(args)?;
```
jbourassa, line 66: I noticed this shouldn't been here as I wrote the doc for it.

ianks (approved):
👏🏻 👏🏻 👏🏻 Absolute love this.

Gemfile
```diff
@@ -12,6 +12,8 @@ group :development do
   gem "standard", "~> 1.3"
   gem "get_process_mem"
   gem "ruby-lsp", require: false
+  gem "yard", require: false
+  gem "yard-rustdoc", "~> 0.1", require: false
```
jbourassa, line 16: Not requiring yard because it has a few patches / global additions, we don't need those.

ext/src/ruby_api/func_type.rs
```diff
@@ -26,14 +29,33 @@ impl FuncType {
...
 impl FuncType {
+    /// @yard
+    ///
+    /// A descriptor for a function in a WebAssembly module.
+    /// WebAssembly functions can have 0 or more parameters and results. Each param
+    /// must be a valid WebAssembly type represented as a symbol. The valid symbols are:
+    /// +:i32+, +:i64+, +:f32+, +:f64+, +:v128+, +:funcref+, +:externref+.
+    ///
+    /// @def new(params, results)
+    /// @param params [Array<Symbol>] The function's parameter types
```
saulecabrera, line 40: A small nit; noting it here but applicable widely: can we add some consistency around the ending period? In this case the `param` definition doesn't have one, but there are other cases in which there's one (same for the general description `def` and `see` entries).

ext/src/ruby_api/module.rs
```diff
@@ -19,12 +27,32 @@ impl Module {
...
     }
 
-    pub fn deserialize(engine: &Engine, wat_or_wasm: RString) -> Result<Self, Error> {
-        unsafe { ModuleImpl::deserialize(engine.get(), wat_or_wasm.as_slice()) }
+    /// @yard
+    /// Instantiates a serialized module coming from either {#serialize} or {Wasmtime::Engine#precompile_module}.
+    ///
+    /// The engine serializing and the engine deserializing must:
+    /// * have the same configuration
+    /// * are of the same gem version
```
saulecabrera, line 35:
```suggestion
    /// * be of the same gem version
```

?

ext/src/ruby_api/func_type.rs
jbourassa: Absolutely! I noticed the same as I proofread. I added period everywhere except when it's only ~2 words:

https://github.com/bytecodealliance/wasmtime-rb/blob/0d5644ed7bbf1c368bcfea034e253f380d2f7cdc/ext/src/ruby_api/linker.rs#L79-L80

Arguably these read as `The module name.` and have a period. 🤷 no strong opinion on my end.

jbourassa: For the `@see`, I'd be tempted to **not** end with a period. It gets rendered as a bullet point and it's not really a sentence:

Screenshot 2022-11-11 at 7 00 19 PM

-----


Agreed?

saulecabrera: > I'd be tempted to not end with a period. It gets rendered as a bullet point and it's not really a sentence:

Oh yeah in this case it makes sense, this just denotes my lack of experience with YARD 😄  I'm good with that, generally it's just to enforce some baseline consistency.

saulecabrera (approved):

ext/src/ruby_api/func.rs
```diff
@@ -40,8 +43,26 @@ unsafe impl Sync for ShareableProc {}
...
+    /// @yieldparam *args [Object] Splat of Ruby objects matching the {FuncType}’s params arity.
+    /// @yieldreturn [nil, Object, Array<Object>] The return type depends on {FuncType}’s results arity:
+    ///   * 0 => +nil+
+    ///   * 1 => +Object+
+    ///   * > 1 => +Array<Object>+
+    ///
+    /// @return [Func]
+    ///
+    /// @example
+    ///   store = Wasmtime::Store.new(Wasmtime::Engine.new)
```
sandstrom, line 63: Is this an example of `Func` usage? 

Seems like a copy-paste error from `Store`.

ext/src/ruby_api/func_type.rs
```diff
@@ -26,14 +29,32 @@ impl FuncType {
 }
 
 impl FuncType {
```
sandstrom, line 31: I know this is more of a design comment, than a documentation comment, but does it make sense to expose this as a class in Ruby?

I had expected these to be part of the `Func` class, e.g.

```ruby

my_func = Wasmtime::Func.new(my_store, params: [:i32, :f64], results: [:v128]) do
  # …
end

# and then if introspection is needed, something like this:
my_func.params #=> [:i32, :f64]
my_func.results #=> [:v128]

```

`FuncType` could still be an internal class if needed, but I wouldn't expose it as a part of the public API.

sandstrom (commented):
A few minor comments below:

jbourassa:
The build failed because I introduced a mismatch between a `@def` and `@param` tag. The system is working 👌.


>```
>[warn]: @param tag has unknown parameter name: wat_or_wasm 
>    in file `ext/src/ruby_api/engine.rs' near line 49
>```

ext/src/ruby_api/func.rs
jbourassa: Yes, thanks for catching this. Fixed now:

https://github.com/bytecodealliance/wasmtime-rb/blob/547488dd7e6a8cd218e1714377e490e07cc20f62/ext/src/ruby_api/func.rs#L62-L67

ext/src/ruby_api/func_type.rs
jbourassa: >I know this is more of a design comment (and this is a docs PR)

At this stage, those are the best kind of comments. Don't hold them back!

I agree that it feels redundant in Ruby. The reason I went with it is for consistency with Wasmtime: Rust, C and Python have that representation (but not .NET). There's value in being close to Wasmtime: similar experience across bindings, the canonical Wasmtime doc is more relevant, easier to stay in sync with Wasmtime, etc.

For this particular case though, does it matter? Probably not 🤔.

ext/src/ruby_api/linker.rs
```diff
@@ -32,28 +34,52 @@ impl DataTypeFunctions for Linker {
...
             .borrow_mut()
             .define_unknown_imports_as_traps(module.get())
             .map_err(|e| error!("{}", e))
     }
 
+    /// @yard
+    /// Define an item in this linker.
+    /// @def define(mod, name, item)
+    /// @param mod [String] Module name
+    /// @param name [Module] Import name
```
sandstrom, line 80: Is this an instance of `Module`? Judging by the source it seems to be a string (I've never used Rust though)

(ditto for other methods in this class)

ext/src/ruby_api/memory_type.rs
```diff
@@ -2,13 +2,19 @@ use super::root;
...
 
+/// @yard
+/// @see https://docs.rs/wasmtime/latest/wasmtime/struct.MemoryType.html Wasmtime's Rust doc
 #[derive(Clone, Debug)]
 #[magnus::wrap(class = "Wasmtime::MemoryType")]
 pub struct MemoryType {
     inner: MemoryTypeImpl,
 }
 
 impl MemoryType {
```
sandstrom, line 13: Same thing here. I'd try to make this a private class, and expose the relevant information/API on the `Memory` class instead.

sandstrom (commented):
Two more comments:

ext/src/ruby_api/func_type.rs
sandstrom: I see the reason for it, even though I would lean in the other direction (tune the SDKs/bindings to the host language, i.e. make the Ruby bindings feel more "like ruby").

Maybe a private class would be a good middle ground?

You'd still get mostly 1:1 with Rust, C, Go and Python (from an implementation perspective).

But API-wise, you'd only expose these via `Func` (as in the example above).

If you'd later decide to expose `FuncType` as a public class, the old methods would just be syntactic sugar over using FuncType directly.

***

Somewhat related, I'd do the same thing with `MemoryType` (make it an internal class, expose relevant API directly on `Memory`.

ext/src/ruby_api/linker.rs
jbourassa: Totally. Just fixed it, thanks!

PR #14, opened 2022-11-13
Add  new `Wasmtime::Extern` class

ianks:
Previously, `Extern` was implicitly mapped to the underlying variant. For example, a `Extern::Func` would automatically become `Wasmtime::Func` when loading the extern in Ruby. This differs from `Wasmtime`, where the caller must do the conversion to ensure the export they want is valid.

To accomplish this, I added a `WrappedStruct<T>` helper to codify a bunch of the conversions we were doing previously. This cleans thing up nicely, so [I made a proposal to `magnus` for it](https://github.com/matsadler/magnus/issues/28) as well.

### DIFF
--- a/ext/src/helpers/wrapped_struct.rs
+++ b/ext/src/helpers/wrapped_struct.rs
@@ -13,7 +13,7 @@ pub struct WrappedStruct<T: TypedData> {
 impl<T: TypedData> Clone for WrappedStruct<T> {
     fn clone(&self) -> Self {
         Self {
-            inner: self.inner.clone(),
+            inner: self.inner,
             phantom: PhantomData,
         }
     }
@@ -36,7 +36,7 @@ impl<T: TypedData> WrappedStruct<T> {
     }
 }
 
-impl<'t, T: TypedData> From<WrappedStruct<T>> for Value {
+impl<T: TypedData> From<WrappedStruct<T>> for Value {
     fn from(wrapped_struct: WrappedStruct<T>) -> Self {
         wrapped_struct.to_value()
     }
@@ -53,7 +53,7 @@ impl<T: TypedData> Deref for WrappedStruct<T> {
 impl<T: TypedData> From<T> for WrappedStruct<T> {
     fn from(t: T) -> Self {
         Self {
-            inner: RTypedData::wrap(t).into(),
+            inner: RTypedData::wrap(t),
             phantom: PhantomData,
         }
     }
@@ -76,7 +76,7 @@ where
         })?;
 
         Ok(Self {
-            inner: inner.into(),
+            inner,
             phantom: PhantomData,
         })
     }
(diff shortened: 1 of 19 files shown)


ext/src/helpers/wrapped_struct.rs
```diff
@@ -0,0 +1,83 @@
...
+                ),
+            )
+        })?;
+
+        Ok(Self {
+            inner: inner.into(),
+            phantom: PhantomData,
+        })
+    }
+}
```
jbourassa, line 83: 👌 love this.

jbourassa (commented):
> This differs from `Wasmtime`, where the caller must do the conversion to ensure the export they want is valid.

I also want to stay close to Wasmtime's API (see https://github.com/bytecodealliance/wasmtime-rb/pull/13#discussion_r1020916809), but in this particular case I didn't see the benefits. Can you expand on why you think the wrapper approach is preferable?

I assumed this was done this way in Wasmtime because of static typing, not ergonomics. The no wrapper version feels more idiomatic Ruby to me. For what it's worth, the [Python bindings also return the extern directly](https://github.com/bytecodealliance/wasmtime-py/blob/6dee92c510370da5259e9565672ce453c1167105/wasmtime/_linker.py#L186), without any wrapper.

ianks:
> > This differs from `Wasmtime`, where the caller must do the conversion to ensure the export they want is valid.
> 
> I also want to stay close to Wasmtime's API (see [#13 (comment)](https://github.com/bytecodealliance/wasmtime-rb/pull/13#discussion_r1020916809)), but in this particular case I didn't see the benefits. Can you expand on why you think the wrapper approach is preferable?
> 
> I assumed this was done this way in Wasmtime because of static typing, not ergonomics. The no wrapper version feels more idiomatic Ruby to me. For what it's worth, the [Python bindings also return the extern directly](https://github.com/bytecodealliance/wasmtime-py/blob/6dee92c510370da5259e9565672ce453c1167105/wasmtime/_linker.py#L186), without any wrapper.

My thinking is that the implicit behavior could cause some strange, potentially undesirable behavior. For example, both `Global` and `Table` expose `get` and `set` methods. There may be unintended consequences of calling `set` on a `Global`, on what you assumed to have been a `table`.

There's also `Memory` and `SharedMemory`, and potentially other overlapping interfaces... By going through an `Extern`, we ensure the user indicates what they expect the underlying extern type to be. So in this case, my intuition is to default to the "correct" rather than "slightly more convenient." If this is burdensome to users, it's easy enough to add a helper to do the implicit unwrap thing. Not so easy to go the other way...

jbourassa:
> My thinking is that the implicit behavior could cause some strange, potentially undesirable behavior.

I think that makes sense. The price to pay is quite small: typing `to_...`, and 1 extra object allocation. :shipit: and I'll rebase #2 on top of it.

ext/src/ruby_api/instance.rs
```diff
@@ -89,15 +93,15 @@ impl Instance {
     /// @def exports
     /// @return [Hash{String => Func, Memory}]
```
jbourassa, line 94: We're gonna have to update that signature, along with `Instance#export` and `Linker#get`.

I can take care of it once it's merged though.

jbourassa (approved):

PR #26, opened 2022-11-14
Add new `mem:check` task to run Valgrind in CI

ianks:
This PR integrates the great [`ruby_memcheck`](https://github.com/Shopify/ruby_memcheck) gem into the repo. It's super powerful... I immediately [found a memory leak in magnus](https://github.com/matsadler/magnus/pull/30) and was able to fix it. 

As of this PR, there are no leaks detected, so we should be good to integrate into CI. One question I have is: Should we run it on every PR, or no? I'm thinking that might be too much, but we should at least run it on `main`. Thoughts?

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -41,6 +41,7 @@ jobs:
       - name: Run "mem:check" task
         env:
           RSPEC_FORMATTER: "progress"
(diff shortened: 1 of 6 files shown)


jbourassa (approved):
> One question I have is: Should we run it on every PR, or no?

Given it's relatively fast, there's no harm in running it, no? Are you thinking it would be flaky?

PR #28, opened 2022-11-16
Improve the docs after #14

jbourassa:
- Bump yard-rustdoc for correctly tagging `rb_self` as instance methods and allowing docs on `Enum`.
- Define `ConversionError` in Ruby so that YARD can see it. yard-rustdoc does not support inheritance nor declaring classes without a Struct / Enum, so defining it in Ruby was much easier than changing yard-rustdoc to support this.
- Update the documentation for Extern after #14

### DIFF
--- a/ext/src/ruby_api/externals.rs
+++ b/ext/src/ruby_api/externals.rs
@@ -5,7 +5,7 @@ use magnus::{
 };
 
 /// @yard
-/// An external item to a WebAssembly module, or a list of what can possibly be exported from a wasm module.
+/// An external item to a WebAssembly module, or a list of what can possibly be exported from a Wasm module.
 /// @see https://docs.rs/wasmtime/latest/wasmtime/enum.Extern.html Wasmtime's Rust doc
 #[derive(TypedData)]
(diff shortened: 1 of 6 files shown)


ext/src/ruby_api/externals.rs
```diff
@@ -5,7 +5,7 @@ use magnus::{
 };
 
 /// @yard
-/// An external item to a WebAssembly module, or a list of what can possibly be exported from a wasm module.
+/// An external item to a WebAssembly module, or a list of what can possibly be exported from a Wasm module.
```
jbourassa, line 8: Nit: we use Wasm everywhere else, so I updated it for consistency.

ext/src/ruby_api/externals.rs
```diff
@@ -25,7 +25,9 @@ impl DataTypeFunctions for Extern {
 
 impl Extern {
     /// @yard
-    /// @return [Wasmtime::Func] The exported function, if this is a function.
```
jbourassa, line 28: Nit: we refer to Wasmtime classes elsewhere without the namespace, updated for consistency.

ianks (approved):

PR #30, opened 2022-11-17
Implement Trap

### LINKED ISSUE #24: Raise `Trap` exception for Wasm traps
Any Wasm execution failure will raise a `Wasmtime::Exception`, but they should be a distinct, more precise exception. I suggest we instead raise `Trap`, which inherit from `Wasmtime::Exception`.

jbourassa:
Closes #24

This PR is (mainly) split in 2 commits:
1. Refactor code and tests so that Wasm error handling happens in 1 place
2. Add `Trap` exception

The perfect solution for defining enums would've had the following properties:
- Self-documenting
- Protect against typos (e.g. not just symbols)
- Can be used in with `case` (e.g. not just methods)
- Easy to keep in sync between Rust and Ruby

This is the closest I could come that ticked most of these boxes. The
current solution's downside is that if you add a trap code, you need
to go and update the Ruby file too.

---

I decided not to implement `Trap#trace` and the frames stuff because (1) it's a net addition (2) I don't have a use case for it yet and (3) it isn't the most pressing thing.

### DIFF
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -14,3 +14,4 @@ lazy_static = "1.4.0"
 magnus = { git = "https://github.com/matsadler/magnus", features = ["rb-sys-interop"] }
 rb-sys = "~0.9.38"
 wasmtime = "2.0.2"
+anyhow = "*" # Use whatever Wasmtime uses
--- a/ext/src/helpers/wrapped_struct.rs
+++ b/ext/src/helpers/wrapped_struct.rs
@@ -18,6 +18,7 @@ impl<T: TypedData> Clone for WrappedStruct<T> {
         }
     }
 }
+impl<T: TypedData> Copy for WrappedStruct<T> {}
 
 impl<T: TypedData> WrappedStruct<T> {
     /// Gets the underlying struct.
@@ -26,7 +27,7 @@ impl<T: TypedData> WrappedStruct<T> {
     }
 
     /// Get the Ruby [`Value`] for this struct.
-    pub fn to_value(&self) -> Value {
+    pub fn to_value(self) -> Value {
         self.inner.into()
     }
 
--- a/lib/wasmtime/trap_code.rb
+++ b/lib/wasmtime/trap_code.rb
@@ -1,0 +1,16 @@
+module Wasmtime
+  module TrapCode
+    STACK_OVERFLOW = :stack_overflow
+    HEAP_MISALIGNED = :heap_misaligned
+    TABLE_OUT_OF_BOUNDS = :table_out_of_bounds
+    INDIRECT_CALL_TO_NULL = :indirect_call_to_null
+    BAD_SIGNATURE = :bad_signature
+    INTEGER_OVERFLOW = :integer_overflow
+    INTEGER_DIVISION_BY_ZERO = :integer_division_by_zero
+    BAD_CONVERSION_TO_INTEGER = :bad_conversion_to_integer
+    UNREACHABLE_CODE_REACHED = :unreachable_code_reached
+    INTERRUPT = :interrupt
+    ALWAYS_TRAP_ADAPTER = :always_trap_adapter
+    UNKNOWN = :unknown
+  end
+end
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -444,6 +444,7 @@ dependencies = [
 name = "ext"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "lazy_static",
  "magnus",
  "rb-sys",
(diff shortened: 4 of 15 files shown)


jbourassa (commented):
The `mem:check` task found issues:

>```
> ==11329== valgrind: Unrecognised instruction at address 0x5829004.
> ==11329== Your program just tried to execute an instruction that > Valgrind
> ==11329== did not recognise.  There are two possible reasons for this.
> ==11329== 1. Your program has a bug and erroneously jumped to a non-code
> ==11329==    location.  If you are running Memcheck and you just saw a
> ==11329==    warning about a bad jump, it's probably your program's fault.
> ==11329== 2. The instruction is legitimate but Valgrind doesn't handle it,
> ==11329==    i.e. it's Valgrind's fault.  If you think this is the case or
> ==11329==    you are not sure, please let us know and we'll try to fix it.
> ==11329== Either way, Valgrind will now raise a SIGILL signal which will
> ==11329== probably kill your program.
>```

After tracking this down, I found out that it happens in Wasmtime on trap: a pure Rust program running a Wasm module that traps will trigger the same error. Note that it also happens on the latest valgrind, I tried a fresh copy from source (3.20)

I'm not sure what to do with this besides filing an issue on Wasmtime and moving on.

ext/src/ruby_api/errors.rs
```diff
@@ -48,6 +49,17 @@ macro_rules! conversion_err {
     };
 }
 
+pub fn handle_wasm_error(store: &Store, error: anyhow::Error) -> Error {
```
saulecabrera, line 52: Any particular reason why this isn't part of the store?

saulecabrera: Wanted to try this out, but `rake` commands keep failing (also I didn't dig deep enough on the cause)

jbourassa: > Any particular reason why this isn't part of the store?

Hmm, not really. I guess it can be! Once we merge #2 this will be StoreContextValue, but not a big deal: we can move the method.

ext/Cargo.toml
```diff
@@ -14,3 +14,4 @@ lazy_static = "1.4.0"
 magnus = { git = "https://github.com/matsadler/magnus", features = ["rb-sys-interop"] }
 rb-sys = "~0.9.38"
 wasmtime = "2.0.2"
+anyhow = "1.*.*"
```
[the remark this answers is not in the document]
jbourassa: I didn't know about `anyhow = "1"`. I can update the PR to it. The intent is: use whatever Wasmtime uses, period.

Could we make it `anyhow = "*"` to guarantee that it picks Wasmtime's version?

ext/src/ruby_api/errors.rs
saulecabrera: I was just thinking that it could be nicer, since the only thing that you'd need to pass in is the error.

jbourassa: Done!

ext/src/helpers/wrapped_struct.rs
```diff
@@ -18,6 +18,7 @@ impl<T: TypedData> Clone for WrappedStruct<T> {
         }
     }
 }
+impl<T: TypedData> Copy for WrappedStruct<T> {}
```
jbourassa, line 21: Don't see a reason why not, under the (transparent) wrappers it's basically a `VALUE`.

ext/src/ruby_api/errors.rs
```diff
@@ -48,6 +49,17 @@ macro_rules! conversion_err {
...
+    store
+        .context_mut()
+        .data_mut()
+        .take_last_error()
```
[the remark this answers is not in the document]
jbourassa: Agreed, PR up: https://github.com/bytecodealliance/wasmtime-rb/pull/31

jbourassa:
Re the [`mem:check` / valgrind issue](https://github.com/bytecodealliance/wasmtime-rb/pull/30#pullrequestreview-1183549106): it looks like a false positive, a very similar (or exactly the same?) issue had already been reported to Wasmtime: https://github.com/bytecodealliance/wasmtime/issues/3061. In that report, the issue is on `ud2`. In our case, we don't actually see what the instruction is, not sure why.

lib/wasmtime/trap_code.rb
```diff
@@ -0,0 +1,16 @@
+module Wasmtime
+  module TrapCode
```
jbourassa, line 2: Pulled this out of `Trap` and into its own module for symmetry with Wasmtime.

saulecabrera (approved):

PR #32, opened 2022-11-19
Prepare for prerelease

jbourassa:
Prepare for a pre-release to RubyGems:

- Rename the gem to `wasmtime` (no `-rb`)

- Change gemspec author to BytecodeAlliance
  This is what [Wasmtime](https://github.com/bytecodealliance/wasmtime/blob/54cfa4df3418961b22a6d5c16a2954b107341080/Cargo.toml#L104) and the Python bindings do.

- Add release workflow
  I developped the workflow against a private fork -- [`wasmtime-jb`](https://rubygems.org/gems/wasmtime-jb/versions). I was testing in a simplified environment but should be close enough that it'll work here 🤞.

- Bump version
  ~I picked `2.0.2-alpha1` given the [versioning policy is that bindings should map to a Wasmtime release](https://bytecodealliance.org/articles/wasmtime-1-0-fast-safe-and-production-ready#:~:text=For%20example%2C%20if%20you%20use%20wasmtime%2Dpy%207.0%2C%20you%20can%20be%20sure%20that%20you%E2%80%99re%20using%20Wasmtime%207.0.).~
  After chatting with Saúl, we chose `0.3.0` so that when we release a "stable" version we can start following Wasmtime's normal versioning scheme.

- Add CONTRIBUTING.md
  For now that's mostly to document the release process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,28 @@
+# Contributing
+
+## Getting started
+
+Install dependencies:
+
+```
+bundle install
+```
+
+Compile the gem, run the tests & Ruby linter:
+
+```
+bundle exec rake
+```
+
+## Releasing
+
+1. Bump the `VERSION` in `lib/wasmtime/version.rb`
+1. Create a new tag for that release, prefixed with `v` (`git tag v1.0.0`):
+  
+  ```
+  git tag v$(grep VERSION lib/wasmtime/version.rb | head -n 1 | cut -d'"' -f2)
+  git push --tags
+  ```
+1. The release workflow will run and push a new version to RubyGems and create
+   a new draft release on GitHub. Edit the release notes if needed, then
+   mark the release as published when the release workflow succeeds.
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Releasing
+
+1. Bump the `VERSION` in `lib/wasmtime/version.rb`
+1. Create a new tag for that release, prefixed with `v` (`git tag v1.0.0`):
+  
+  ```
+  git tag v$(grep VERSION lib/wasmtime/version.rb | head -n 1 | cut -d'"' -f2)
+  git push --tags
```
saulecabrera, line 24: Not for this PR, I wonder is there's any value in having a small script to automate this.

saulecabrera (approved):

CONTRIBUTING.md
jbourassa: Honestly if feels so small that I don't think it's worth it _yet_. But 👀 [rb-sys has a nice bump release script](https://github.com/oxidize-rb/rb-sys/blob/main/rakelib/release.rake#L2-L17), maybe we could do something similar eventually:
1. ask for the new version
1. update refs in the code
1. commit changes
1. create the tag
1. print "if that all looks good to you, `git push --tags`"

PR #43, opened 2022-11-24
Support WASI

### LINKED ISSUE #21: Support WASI
Add support for WASI. The [C API](https://github.com/bytecodealliance/wasmtime/blob/main/crates/c-api/src/wasi.rs) and [Python bindings](https://github.com/bytecodealliance/wasmtime-py/blob/main/wasmtime/_wasi.py) can be used as starting points. It would also be great if we could support stdin and stdout without using files (binary String, maybe?).

jbourassa:
Fixes #21.

Add APIs to support WASI:

- ~`Linker#define_wasi` which... defines WASI for a linker.~ This is now a bool kwarg on `Linker.new`: `Linker.new(engine, wasi: true)`
- `WasiConfig`: a builder object that holds a WASI configuration and does nothing else. The config can be changed at will -- even by mutating the underlying Ruby objects -- and and re-used across stores.
- ~`Store#configure_wasi` to set the WASI context from `WasiConfig`. Once that's done, future changes to `WasiConfig` won't trickle to the store: the WASI state has been copied to the store.~ This is now a `wasi_ctx` keyword argument on `Store.new`
- `WasiExit`: exception for WASI's `proc_exit`.

This more or less follows the pattern set by the C API, except:
- No "inherit env" and "inherit args". I initially had it in, but it segfaulted when reading `std::env::args()`. I don't see a lot of value in that feature, it's ambiguous (should we use ARGV which can be mutated, or the original program's args?), and can easily be replicated by sending in `ARGV` / `ENV`.
- In C, setting the WASI context is `wasmtime_context_set_wasi(wasi_config_t)`. I chose `Store#configure_wasi` for symmetry with `Linker#define_wasi`. Also, `configure_wasi` materializes the Builder object into a WASI context, which can do file operation (opening files for stdin/out/err) -- not what I'd expect a "setter" to do.
- On `Linker#instantiate`, we raise if the the store does not have a WASI context. This seems nicer than `thread '<unnamed>' panicked at...` in Ruby. I'm also not confident that catching the panic would leave the program in a good state (but probably?).

^ Feedback / thoughts welcome on everything, but on those last 2 points in particular.

----

In a future PR, I'll introduce reading writeable streams (stdout, stderr) as Ruby strings.


ext/src/ruby_api/store.rs
```diff
@@ -1,17 +1,20 @@
...
 use magnus::{
     exception::Exception, function, method, scan_args, value::BoxValue, DataTypeFunctions, Error,
     Module, Object, TypedData, Value, QNIL,
 };
 use std::cell::{RefCell, UnsafeCell};
 use std::convert::TryFrom;
 use wasmtime::{AsContext, AsContextMut, Store as StoreImpl, StoreContext, StoreContextMut};
+use wasmtime_wasi::{I32Exit, WasiCtx};
 
-#[derive(Debug)]
```
jbourassa, line 11: Had to drop this because `WasiCtx` does not implement `Debug`.

examples/wasi.rb
```diff
@@ -0,0 +1,19 @@
+require "wasmtime"
+
+engine = Wasmtime::Engine.new
+mod = Wasmtime::Module.from_file(engine, "spec/fixtures/wasi-debug.wasm")
+
+linker = Wasmtime::Linker.new(engine)
+linker.define_wasi
```
saulecabrera, line 7: I started having second thoughts about this, and I've been thinking that it might better to have two constructors, one for the default case and one with WASI:
```rb
Wasmtime::Linker::new # default
Wasmtime::Linker::with_wasi 
```

I can't of a reason in which you'd need to call `define_wasi` at a later point in time after creating the linker (or maybe we discussed this and I just don't recall)

ext/src/ruby_api/store.rs
```diff
@@ -102,6 +114,20 @@ impl Store {
         self.context().data().user_data()
     }
 
+    /// @yard
+    /// Configure the WASI context on the store.
+    /// @def configure_wasi(wasi_config)
+    /// @param wasi_config [WasiConfig]
+    /// @return [Store] +self+
+    pub fn configure_wasi(
```
saulecabrera, line 122: If we want to be more explicit about what's going on under the hood maybe we can name this `set_wasi_from_config`.

ext/src/ruby_api/wasi_config.rs
```diff
@@ -0,0 +1,283 @@
...
+#[derive(Default)]
+struct WasiConfigInner {
+    stdin: Option<ReadStream>,
+    stdout: Option<WriteStream>,
+    stderr: Option<WriteStream>,
+    env: Option<RHash>,
+    args: Option<RArray>,
+}
+
+macro_rules! mark_some {
```
saulecabrera, line 50: This use-case doesn't seem to warrant a macro, especially since it only appears to be used in this module, or I might be wrong and we could reuse this elsewhere? Can this be replaced by an `iterator` + `for_each`?

ext/src/ruby_api/wasi_config.rs
```diff
@@ -0,0 +1,283 @@
...
+
+impl DataTypeFunctions for WasiConfig {
+    fn mark(&self) {
+        self.inner.borrow().mark();
+    }
+}
+
+type RbSelf = WrappedStruct<WasiConfig>;
+
+impl WasiConfig {
```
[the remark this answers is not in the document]
saulecabrera: The other question that comes to mind here, is why did we decided to go with a builder patter rather than having plain setters?

ext/src/ruby_api/wasi_config.rs
saulecabrera: No strong opinions there, just personally I like to avoid them when possible since they introduce complexity, especially for newcomers.

jbourassa: > we could reuse this elsewhere

This feels oddly specific, I don't think we'd use it elsewhere. At least not as written.

🤷 initially I had 5 of these so I thought it made it neater. But with only 3, I agree it's not super valuable. 

> Can this be replaced by an `iterator` + `for_each`?

I'm not sure how to go about that in a efficient manner (and not sure it's worth it?). Maybe we can explore that in our next pairing session, until then I'll remove the macro.

ext/src/ruby_api/wasi_config.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// @yard
+    /// Set the arguments (argv) to the specified +Array+.
+    /// @param args [Array<String>]
+    /// @def set_argv(args)
+    /// @return [WasiConfig] +self+
+    pub fn set_argv(rb_self: RbSelf, argv: RArray) -> Result<RbSelf, Error> {
+        let mut inner = rb_self.get()?.inner.borrow_mut();
+        inner.args = Some(argv);
+        Ok(rb_self)
+    }
```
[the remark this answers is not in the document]
jbourassa: >What do you find ambiguous from inherit args and inherit env?

In Ruby, you can mutate the `ENV` and `ARGV`. [Changes to `ENV` are propagated to the process' env](https://github.com/ruby/ruby/blob/4491bb740a9506d76391ac44bb2fe6e483fec952/hash.c#L7261-L7265), but `ARGV` aren't. It's ambiguous (to me) which is the expected behaviour: should we use the Ruby-mutated ARGV, or should we use the original args (which segfaults with Rust' implementation, oops!).

By not providing the feature and letting integrators inject ENV & ARGV, we have a smaller API surface and remove the ambiguity.


-----


> Could we provide an api that simulates the inheriting process by ready from `ARGV` / `ENV` under the hood?

Yeah, totally, it'd be quite easy. I didn't see a lot of value in doing it though, as per the above.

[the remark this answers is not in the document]
jbourassa: > Also should the explicit setters be named `args` and `envs`? My only reason for suggesting that is that they map directly to how that API is surface in `WasiCtxBuilder`.

I took example on the C WASI API which uses [`wasi_config_set_argv`](https://github.com/bytecodealliance/wasmtime/blob/75cd888e232aad18ac1c7bd0f40a6929ffba30c6/crates/c-api/src/wasi.rs#L106) and [`wasi_config_set_env`](https://github.com/bytecodealliance/wasmtime/blob/75cd888e232aad18ac1c7bd0f40a6929ffba30c6/crates/c-api/src/wasi.rs#L125).

We borrow pieces from the Rust and C APIs. Similar to Rust's `WasiCtxBuilder`, `WasiConfig` has a builder-style API. Similar to the C API's `wasi_config_t`, `WasiConfig` holds a configuration which is later applied.

Weakly-held opinion: `env` and `argv` is slightly better because it maps more closely to the Ruby constants, and there's a precedent in the C API anyways. Thoughts?

examples/wasi.rb
saulecabrera: > What's your reasoning for preferring a separate constructor?

WASI is so special and so commonly used that it seems to me that having two constructors is justified. It also feels way more solid than having to call a second method to define something that could be bundled from the start, reducing the risk of not setting up a WASI based runtime correctly. 

> By using a dedicated constructor, we're essentially making 2 types of linkers: WASI and non-WASI. That would prevent another similar concept from being used together with WASI

I see it as if you're creating a linker that automatically bundles WASI imports; but I'm not convinced that this addition prevents adding functionality in the future. I think that if we were optimizing for the extensibility use-case, an argument could be made that we shouldn't even be defining `define_wasi` in the Linker itself.

ext/src/ruby_api/wasi_config.rs
saulecabrera: > I'm not sure how to go about that in a efficient manner (and not sure it's worth it?)

Not worth it IMO, I added it as an option in case you still want to avoid writing all the statements by hand.

ext/src/ruby_api/wasi_config.rs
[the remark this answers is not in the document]
jbourassa: > What's the reason to deviate from `WasiCtx`?

As I alluded to in another comment, this `WasiConfig` is more like the C API's `wasi_config_t`. It lets you describe a WASI context once, and we'll create the `WasiCtx` for the config when needed. That's why I stuck to the "wasi config" name.

The Rust `WasiCtx` is what you get after "applying" (for lack of a better word) the Ruby `WasiConfig` to a Ruby `Store`. `WasiCtx` is already configured with random, clock, etc., the `WasiConfig` is not. We could eventually add methods to `WasiConfig` so it can build a deterministic wasi context, or maybe even a `WasiCtx` that's backed by Ruby Procs (although that's possibly slower).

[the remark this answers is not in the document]
jbourassa: > why did we decided to go with a builder patter rather than having plain setters?

I saw that `WasiCtxBuilder` uses a builder pattern and thought that was pretty elegant. It can remove the need to repeat a variable name over and over.

Also, when I think of setters, I think of simple property assigns. In this case, some settings are mutually exclusive (as expressed with the enums), thus setting a field can have effect on other fields. The current builder API does not make this obvious, but it still feels better than setters. We could make this clearer, at the cost of extra boilerplate, making the API more complex, and extra object allocations (off the hot path):

```ruby
# With extra object:
Wastime::WasiConfig.new
  .stdout(Wastime::WasiConfig::File.new("tmp/stdout"))
  .stdin(Wasmtime::WasiConfig::String.new("stdin bytes"))

# With blocks:
Wastime::WasiConfig.new
  .stdout { |stdout| stdout.file("tmp/stdout") }
  .stdin { |stdin| stdin.string("stdin bytes") }
```

I like these APIs, but they feel too complex for the task at hand.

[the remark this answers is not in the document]
saulecabrera: > The Rust WasiCtx is what you get after "applying" (for lack of a better word) the Ruby WasiConfig 

I should've been more clear here. I meant why we deviated from the `WasiCtx` and `WasiCtxBuilder` naming. Your explanation above makes sense, but I'm not sure that the C-API should be treated as "canonical", instead I think that our canonical should be the Rust one where possible. That's why I think that using `WasiCtx` should be preferred. 

I'm good with the setters part, I was asking more out of curiosity.

ext/src/ruby_api/wasi_config.rs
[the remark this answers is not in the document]
saulecabrera: > n Ruby, you can mutate the ENV and ARGV. Changes to ENV are propagated to the process' env, but ARGV aren't. It's ambiguous (to me) which is the expected behaviour: should we use the Ruby-mutated ARGV, or should we use the original args (which segfaults with Rust' implementation, oops!).

> By not providing the feature and letting integrators inject ENV & ARGV, we have a smaller API surface and remove the ambiguity.


I assumed this was referring to the `Rust` api or api as a whole. This makes sense to me. Keeping the `set_<argv|env>` makes sense.

jbourassa:
@saulecabrera I think all comments are addressed. I've made the changes as additional commits so they're easy to review. The simplest way to see the resulting API is [the example](https://github.com/bytecodealliance/wasmtime-rb/pull/43/files#diff-20338b850a0fe56d4dd9192f0178cf45dce348980604c46160dfbc026bc6b5c5) or [the tests](https://github.com/bytecodealliance/wasmtime-rb/pull/43/files#diff-1e9e4a10c2047ae52d962e46b084f6b695c0c613ceac71bbc980d2cc5bbd51ee). Let me know what you think!

saulecabrera (approved):

PR #44, opened 2022-11-24
Fix typo in readme example

sandstrom:
Remove stray parenthesis (I don't think it should be there?)


jbourassa:
Yeah, HEREDOC is a bit weird. The comment fixes are relevant though, I'll incorporate those. Thanks.

PR #49, opened 2022-11-29
Add custom task to build the source gem

ianks:
This PR adds a new `pkg:ruby` Rake task to bundle a source-code-only version of the gem. This gem is intended for users who cannot use the precompiled gems (i.e. Ruby head, or when `force_ruby_platform` is used with bundler).

Before packaging the gem, this PR vendors the crates necessary for compilation so `gem install` can be done offline, and avoids hitting the cargo registry or github.

### DIFF
--- a/rakelib/doc.rake
+++ b/rakelib/doc.rake
@@ -1,5 +1,9 @@
 require "yard/rake/yardoc_task"
 
+CLOBBER.include("doc")
+CLEAN.include(".yardoc")
+CLEAN.include("tmp/doc")
+
 YARD::Rake::YardocTask.new do |t|
   t.options += ["--fail-on-warn"]
 
@@ -50,19 +54,16 @@ namespace :doc do
 
   desc "Generate Rust documentation as JSON"
   task :rustdoc do
-    run(<<~CMD)
+    sh <<~CMD
       cargo +nightly rustdoc \
-        --target-dir tmp \
+        --target-dir tmp/doc/target \
         -p ext \
         -- -Zunstable-options --output-format json \
         --document-private-items
     CMD
-  end
 
-  def run(cmd)
-    system(cmd)
-    fail if $? != 0
+    cp "tmp/doc/target/doc/ext.json", "tmp/doc/ext.json"
   end
 end
 
-task doc: "doc:default"
+task doc: ["compile", "doc:default"]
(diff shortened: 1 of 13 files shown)


rakelib/spec.rake
```diff
@@ -0,0 +1,9 @@
+CLEAN.include(".rspec_status")
```
ianks, line 1: Took the opportunity to extract this file...

rakelib/pkg.rake
```diff
@@ -0,0 +1,67 @@
...
+      cargo_config_path = ".cargo/config"
+      final_gemspec = Gem::Specification.load(File.basename(gemspec_path))
+
+      puts "Vendoring cargo dependencies to #{cargo_config_path}..."
+      mkdir_p ".cargo"
+      sh "cargo vendor --versioned-dirs --locked #{vendor_dir} >> #{cargo_config_path} 2>/dev/null"
+
+      final_gemspec.files += dirglob("**/.cargo/**/*")
+      final_gemspec.files += dirglob("./#{vendor_dir}/**/*")
+      final_gemspec.files.reject! { |f| File.directory?(f) }
```
jbourassa, line 43: Nit: first time I ran this, it didn't complete because I had empty dirs from some prior exploration and that seemed to break the script. Somehow the gemspec ends up including files that do not exist, not sure why:

```
Gem::InvalidSpecificationException: ["ext/wasmtime"] are not files
wasmtime-rb/rakelib/pkg.rake:46:in `block (3 levels) in <top (required)>'
```

Removing the empty dir fixed it for me.  If you want to repro: `mkdir -p ext/wasmtime && rake pkg:ruby`

rakelib/doc.rake
```diff
@@ -50,19 +54,16 @@ namespace :doc do
 
   desc "Generate Rust documentation as JSON"
   task :rustdoc do
-    run(<<~CMD)
+    sh <<~CMD
```
jbourassa, line 57: TIL [`sh`](https://www.rubydoc.info/gems/rake/FileUtils:sh), neat.

jbourassa (approved):
Thanks @ianks! Once that's in, we should replace this step of the release workflow:


https://github.com/bytecodealliance/wasmtime-rb/blob/a34ed64694811bf88606c5bd87f5211bf603a293/.github/workflows/release.yml#L83-L86

rakelib/pkg.rake
ianks: solved in https://github.com/bytecodealliance/wasmtime-rb/pull/49/commits/08ef238cce08418af9853d467f3947ea2ae2f16a

PR #50, opened 2022-11-29
Add fuel

### LINKED ISSUE #25: Support fuel
Engine, Store and Caller do not expose fuel-related methods.

jbourassa:
Add fuel:
- Allow configuring an Engine to use fuel
- Add fuel methods to Store and Caller
- Add trap code for out of fuel

Fixes #25

### DIFF
--- a/ext/src/ruby_api/config.rs
+++ b/ext/src/ruby_api/config.rs
@@ -26,6 +26,13 @@ impl Config {
     }
 
     /// @yard
+    /// @def consume_fuel=(enabled)
+    /// @param enabled [Boolean]
+    pub fn set_consume_fuel(&self, enabled: bool) {
+        self.inner.borrow_mut().consume_fuel(enabled);
+    }
+
+    /// @yard
     /// @def epoch_interruption=(enabled)
     /// @param enabled [Boolean]
     pub fn set_epoch_interruption(&self, enabled: bool) {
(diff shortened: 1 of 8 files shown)


ext/src/ruby_api/config.rs
```diff
@@ -52,6 +59,8 @@ pub fn init() -> Result<(), Error> {
 
     class.define_singleton_method("new", function!(Config::new, 0))?;
 
+    class.define_method("consume_fuel=", method!(Config::set_consume_fuel, 1))?;
+
```
jbourassa, line 63: I'm not a huge fan of these setters, I think a builder-like API would be nice here too, and coherent with `WasiCtxBuilder`. But I decided to not shave this yak just yet.

ext/src/ruby_api/func.rs
```diff
@@ -338,6 +338,30 @@ impl<'a> Caller<'a> {
         }
     }
 
+    /// (see Store#fuel_consumed)
+    /// @yard
```
jbourassa, line 342: Putting `@yard` last so the `(see ...)` is first, which is a requirement of YARD to use ["reference tags"](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md#reference-tags).

I'll fix this in YARD-rustdoc, but figured that workaround was acceptable.

spec/unit/fuel_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      context "on Caller" do
+        it desc do
+          func = Func.new(send(store), FuncType.new([], [])) do |caller|
+            instance_exec(caller, &block)
+          end
+          func.call
+        end
+      end
+    end
```
jbourassa, line 20: This is a little strange but allows writing the test once for both `Store` and `Caller`. I think it's worth it.

saulecabrera (approved):

PR #52, opened 2022-12-01
Add `Wasmtime.wat2wasm`

jbourassa:
This is a minor addition but would be helpful to have as I'm integrating wasmtime-rb in a real-world project. For my use case I need binary Wasm (not WAT) and would like to test using simple, sample programs. Bundling these as WAT in my test files would be easier than having the binary Wasms.

### DIFF
--- a/ext/src/ruby_api/mod.rs
+++ b/ext/src/ruby_api/mod.rs
@@ -2,7 +2,7 @@
 #![allow(rustdoc::invalid_html_tags)]
 #![allow(rustdoc::bare_urls)]
 #![allow(rustdoc::invalid_rust_codeblocks)]
-use magnus::{define_module, memoize, Error, RModule};
+use magnus::{define_module, function, memoize, Error, RModule, RString};
 
 mod config;
 mod convert;
@@ -28,8 +28,27 @@ pub fn root() -> RModule {
     *memoize!(RModule: define_module("Wasmtime").unwrap())
 }
 
+// This Struct is a placeholder for documentation, so that we can hang methods
+// to it and have yard-rustdoc discover them.
+/// @yard
+/// @module
+pub struct Wasmtime;
+impl Wasmtime {
+    /// @yard
+    /// Converts a WAT +String+ into Wasm.
+    /// @param wat [String]
+    /// @def wat2wasm(wat)
+    /// @return [String] The Wasm represented as a binary +String+.
+    pub fn wat2wasm(wat: RString) -> Result<RString, Error> {
+        wat::parse_str(unsafe { wat.as_str()? })
+            .map(|bytes| RString::from_slice(bytes.as_slice()))
+            .map_err(|e| crate::error!("{}", e))
+    }
+}
+
 pub fn init() -> Result<(), Error> {
-    let _ = root();
+    let wasmtime = root();
+    wasmtime.define_module_function("wat2wasm", function!(Wasmtime::wat2wasm, 1))?;
 
     errors::init()?;
     trap::init()?;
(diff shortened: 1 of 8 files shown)


rakelib/doc.rake
```diff
@@ -15,10 +15,10 @@ YARD::Rake::YardocTask.new do |t|
       case yard_object.type
       when :module
         mod = Object.const_get(yard_object.path)
-        errors << "Not a module: #{mod.path}" unless mod.is_a?(::Module)
+        errors << "Not a module: #{mod}" unless mod.is_a?(::Module)
       when :class
         klass = Object.const_get(yard_object.path)
-        errors << "Not a class: #{klass.path}" unless klass.is_a?(::Class)
+        errors << "Not a class: #{klass}" unless klass.is_a?(::Class)
```
jbourassa, line 21: Oops: that never worked, `Class#path` `Module#path` doesn't exist.

ext/src/ruby_api/mod.rs
```diff
@@ -28,8 +28,27 @@ pub fn root() -> RModule {
     *memoize!(RModule: define_module("Wasmtime").unwrap())
 }
 
+// This Struct is a placeholder for documentation, so that we can hang methods
+// to it and have yard-rustdoc discover them.
+/// @yard
+/// @module
+pub struct Wasmtime();
```
saulecabrera, line 35:
```suggestion
pub struct Wasmtime;
```

ext/src/ruby_api/mod.rs
```diff
@@ -28,8 +28,27 @@ pub fn root() -> RModule {
...
+impl Wasmtime {
```
saulecabrera, line 36: Did you consider adding this under a different namespace? Personally I find it confusing to have this under the Wasmtime namespace, mainly because Wasmtime doesn't offer this functionality.

ext/src/ruby_api/mod.rs
jbourassa: Oh, I thought these parens were required. Fixed! 👍

ext/src/ruby_api/mod.rs
jbourassa: Do you mean on the Ruby layer, this struct specifically? Personally it doesn't bother me as it'll end up on the `Wasmtime` constant. Alternatively, we could name this `Root`, but that feels worse.

> Personally I find it confusing to have this under the Wasmtime namespace, mainly because Wasmtime doesn't offer this functionality.

I see your point, it's in a different crate. From the bindings perspective though, it's often bundled with Wasmtime itself:

- Go has [`wasmtime.Wat2Wasm`](https://github.com/bytecodealliance/wasmtime-go/blob/4b3a40c7a33dca1cfa4e06e6f40e06816f7d2421/README.md?plain=1#L75)
- C has [`wasmtime_wat2wasm`](https://github.com/bytecodealliance/wasmtime/blob/d54a27d0ea83090bf1bd46032a2129677f70df0f/crates/c-api/src/wat2wasm.rs)
- .NET has [`Wasmtime.Module.ConvertText`](https://github.com/bytecodealliance/wasmtime-dotnet/blob/992efb0baee78bad1b4f9cc0a8a5b363313a63c9/src/Module.cs#L326)

saulecabrera: .NET approach seems to be the cleanest to me, but since there's a precedent on this I'll leave it to you if you want to change it.

saulecabrera (approved):

ext/src/ruby_api/mod.rs
ianks: Maybe put it in `Wasmtime::Utils`

ianks (approved):

PR #54, opened 2022-12-02
Allow stdin & stdout to write to String

jbourassa:
Allow WASI's stdin and stdout to write to a Ruby String, which can then be read back from the store.

Opening this as a draft PR because I'm not overly satisfied with the current state:
- the API is clunky, getting stdin/out directly from the Store feels odd. Other possible approaches:
  - introducing `Store#wasi_ctx` that returns a WASI context with the 2 methods on it (but I don't know what else it would ever have, and exposing more methods probably means more data copy, thus slight perf cost)
  - taking a (non frozen) string in `#set_stdin_string`, and modifying that string in place.
- There's a missing abstraction for the `(wasi ctx, stdout string, stderr string)` tuple.

### DIFF
--- a/ext/src/ruby_api/store.rs
+++ b/ext/src/ruby_api/store.rs
@@ -4,7 +4,7 @@ use crate::{define_rb_intern, error, helpers::WrappedStruct};
 use magnus::Class;
 use magnus::{
     exception::Exception, function, method, scan_args, value::BoxValue, DataTypeFunctions, Error,
-    Module, Object, TypedData, Value, QNIL,
+    Module, Object, RString, TypedData, Value, QNIL,
 };
 use std::cell::{RefCell, UnsafeCell};
 use std::convert::TryFrom;
@@ -19,6 +19,8 @@ pub struct StoreData {
     user_data: Value,
     host_exception: HostException,
     wasi: Option<WasiCtx>,
+    wasi_stdout_string: Option<RString>,
+    wasi_stderr_string: Option<RString>,
 }
 
 type BoxedException = BoxValue<Exception>;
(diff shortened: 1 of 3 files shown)


ianks:
Curious about the motivation for this. Is the idea to be able to capture logs when running a 3rd party WASM module or something?

jbourassa:
> Is the idea to be able to capture logs when running a 3rd party WASM module or something?

Yes, that's one use case for it. That's what we do with [Shopify Functions](https://shopify.dev/api/functions/reference/webassembly-api). The idea is to avoid using files to capture stdin and stdout, really. A `String` isn't as flexible as a stream, but it's simple and fast, and doesn't preclude a stream-based API later.

There is an [open PR](https://github.com/bytecodealliance/wasmtime/pull/5210) to achieve a similar goal in the C API too.

Feel free to share any thoughts you might have, I'm curious to hear them!

ianks:
Could we have it be an arbitrary Ruby IO? If you want a string, you could then just use StringIO

jbourassa:
> Could we have it be an arbitrary Ruby IO? If you want a string, you could then just use StringIO

Something like this?

```ruby
# Send in any IO object, but keep a reference to it
stdout = StringIO.new
wasi_config = WasiCtxBuilder.new.set_stdout_io(stdout)

run_wasi_module(wasi_config)

# Read it whenever
stdout.rewind
stdout.read
```

I think there has to be an additional copies for this to work, because we have to call back into Ruby (`#write`) with a `RString` -- right? I can give this a try and see if that copy makes a difference, though I suspect it'll be within noise level.

PR #55, opened 2022-12-05
Add instructions about precompiled gems to readme

ianks:
This should help avoid any confusion caused by Bundler, since Bundler will error out when a platform isn't locked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,16 +21,32 @@ The Wasmtime Ruby bindings are still under development, [some features](https://
 
 ## Installation
 
-Install from RubyGems
+Add the `wasmtime` gem to your Gemfile and run `bundle install`:
 
-```shell
+```ruby
+gem "wasmtime"
+```
+
+Alternatively, you can install the gem manually:
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,16 +21,32 @@ The Wasmtime Ruby bindings are still under development, [some features](https://
 
 ## Installation
 
-Install from RubyGems
+Add the `wasmtime` gem to your Gemfile and run `bundle install`:
 
-```shell
+```ruby
+gem "wasmtime"
```
ianks, line 27: This lets bundler handle the locking, and lets dependabot suggest upgrades.

README.md
```diff
@@ -21,16 +21,32 @@ The Wasmtime Ruby bindings are still under development, [some features](https://
...
+```
+
+Alternatively, you can install the gem manually:
```
ianks, line 30: Suggest this secone, since most people use bundler anyway...

jbourassa (approved):
Thanks!

PR #58, opened 2022-12-05
Add rake task to run examples

ianks:
This PR adds some tasks to make make running examples easier:

```
rake examples        # Run all the examples
rake examples:gcd    # Run examples/gcd.rb
rake examples:hello  # Run examples/hello.rb
rake examples:wasi   # Run examples/wasi.rb
```

Currently, the `wasi.rb` example is failing. Would like to run this check in CI, so should we hold off merging until that is fixed?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,9 @@ jobs:
         env:
           GC_STRESS: "true"
 
+      - name: Run examples
+        run: bundle exec rake examples
+
       - name: Lint ruby
         run: bundle exec rake standard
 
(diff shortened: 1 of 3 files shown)


jbourassa (approved):
Thanks. I've been meaning to get the examples in CI, but had not considered a rake task. Good idea!

> Currently, the `wasi.rb` example is failing.

What are you seeing? Seems to work fine on my machine and CI:

```
$bundle exec rake examples:wasi > /dev/null 2>&1 ; echo $?
0
```

It pints the env, argv and stdin to both stdout and stderr so the output is pretty large, but it works.

PR #68, opened 2022-12-06
Make `wasmtime-rb` usable as a Rust crate

ianks:
This PR makes it so `wasmtime-rb` can be used from a Rust crate. This makes it possible to create a Ruby native gem that uses the `ruby_api` that `wasmtime-rb` exposes.


jbourassa (approved):
I don't see a problem with that, but I haven't paid close attention to the crate API, especially not around what should be public or not.

PR #69, opened 2022-12-06
Add env configurations for `dev` and `release`

ianks:
This PR makes it do we explicitly configure build profiles when compiling the extension.

### New tasks

```
rake env:dev          # Sets up environment variables "dev" builds
rake env:release      # Sets up environment variables "release" builds
rake compile:dev      # Compile the extension in "dev" mode
rake compile:release  # Compile the extension in "release" mode
```

### Default behavior

- **Local development**: compile in `dev` mode (so overflow checks are configured and debugging is easier)
- **CI**: compile in `release` mode to make sure things works with that profile

### DIFF
--- a/rakelib/env.rake
+++ b/rakelib/env.rake
@@ -1,0 +1,13 @@
+namespace :env do
+  desc 'Sets up environment variables "dev" builds'
+  task :dev do
+    ENV["RUST_BACKTRACE"] = "1"
+    ENV["WASMTIME_BACKTRACE_DETAILS"] = "1"
+    ENV["RB_SYS_CARGO_PROFILE"] ||= "dev"
+  end
+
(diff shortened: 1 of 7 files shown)


rakelib/env.rake
```diff
@@ -0,0 +1,13 @@
+namespace :env do
+  desc 'Sets up environment variables "dev" builds'
+  task :dev do
+    ENV["RUST_BACKTRACE"] = "1"
+    ENV["WASMTIME_BACKTRACE_DETAILS"] = "1"
+    ENV["RB_SYS_CARGO_PROFILE"] ||= "dev"
```
jbourassa, line 6: I think that made compilation faster in dev, nice.

jbourassa (approved):

PR #70, opened 2022-12-06
Add table support

### LINKED ISSUE #20: Support Tables
Implement support for Table: https://docs.wasmtime.dev/api/wasmtime/struct.Table.html

jbourassa:
Add support for Wasm tables. Fixes #20.

The first 2 commits are refactors:
1. Move `ToValType` to `convert` as we'll need it not only for Funcs.

2. Move `refs` from `Store` to `StoreData` so it's accessible for both `Store` and `Caller`. This is we have a way of GC marking the extern refs we store in the `Table`.
    
    Storing on the (Ruby) `Table` itself won't work because the user can drop their reference to `Table`, but we must still keep marking whatever they stored in the Table otherwise we could segfault or have data corruption.

The other commits implement tables.

### DIFF
--- a/ext/src/ruby_api/convert.rs
+++ b/ext/src/ruby_api/convert.rs
@@ -29,7 +29,7 @@ impl ToRubyValue for Val {
                 None => Ok(magnus::QNIL.into()),
                 Some(eref) => eref
                     .data()
-                    .downcast_ref::<OnStackValue>()
+                    .downcast_ref::<RbExternRef>()
                     .map(|v| v.0)
                     .ok_or_else(|| error!("failed to extract externref")),
             },
@@ -55,25 +55,31 @@ impl ToWasmVal for Value {
             ValType::ExternRef => {
                 let extern_ref_value = match self.is_nil() {
                     true => None,
-                    false => Some(ExternRef::new(OnStackValue::from(*self))),
+                    false => Some(ExternRef::new(RbExternRef::from(*self))),
                 };
 
                 Ok(Val::ExternRef(extern_ref_value))
             }
-            ValType::FuncRef => Ok(Val::FuncRef(Some(*self.try_convert::<&Func>()?.inner()))),
+            ValType::FuncRef => {
+                let func_ref_value = match self.is_nil() {
+                    true => None,
+                    false => Some(*self.try_convert::<&Func>()?.inner()),
+                };
+                Ok(Val::FuncRef(func_ref_value))
+            }
             ValType::V128 => err!("converting from Ruby to v128 not supported"),
         }
     }
 }
 
-struct OnStackValue(Value);
-impl From<Value> for OnStackValue {
+struct RbExternRef(Value);
+impl From<Value> for RbExternRef {
     fn from(v: Value) -> Self {
         Self(v)
     }
 }
-unsafe impl Send for OnStackValue {}
-unsafe impl Sync for OnStackValue {}
+unsafe impl Send for RbExternRef {}
+unsafe impl Sync for RbExternRef {}
 
 pub trait ToExtern {
     fn to_extern(&self) -> Result<wasmtime::Extern, Error>;
--- a/ext/src/ruby_api/convert.rs
+++ b/ext/src/ruby_api/convert.rs
@@ -1,8 +1,18 @@
-use crate::{err, error};
-use magnus::{Error, TypedData, Value};
+use crate::{define_rb_intern, err, error};
+use magnus::{Error, Symbol, TypedData, Value};
 use wasmtime::{ExternRef, Val, ValType};
 
 use super::{func::Func, memory::Memory, store::StoreContextValue};
+
+define_rb_intern!(
+    I32 => "i32",
+    I64 => "i64",
+    F32 => "f32",
+    F64 => "f64",
+    V128 => "v128",
+    FUNCREF => "funcref",
+    EXTERNREF => "externref",
+);
 
 pub trait ToRubyValue {
     fn to_ruby_value(&self, store: &StoreContextValue) -> Result<Value, Error>;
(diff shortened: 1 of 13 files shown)


ext/src/ruby_api/convert.rs
```diff
@@ -45,25 +55,31 @@ impl ToWasmVal for Value {
...
+                    false => Some(*self.try_convert::<&Func>()?.inner()),
+                };
+                Ok(Val::FuncRef(func_ref_value))
+            }
             ValType::V128 => err!("converting from Ruby to v128 not supported"),
         }
     }
 }
 
-struct OnStackValue(Value);
```
jbourassa, line 59: Renamed this because externrefs aren't necessarily on the stack anymore with table.

ext/src/ruby_api/memory.rs
```diff
@@ -29,6 +29,7 @@ unsafe impl TypedData for Memory<'_> {
         memoize!(magnus::DataType: {
             let mut builder = DataTypeBuilder::<Memory<'_>>::new("Wasmtime::Memory");
             builder.free_immediately();
+            builder.mark();
```
jbourassa, line 32: Unrelated to tables, but I noticed we weren't marking Memory.

jbourassa:
@saulecabrera let me know if you find this hard to review, I can split up the first 2 commits in a separate PR.

ext/src/ruby_api/convert.rs
```diff
@@ -45,25 +55,31 @@ impl ToWasmVal for Value {
...
-impl From<Value> for OnStackValue {
+struct RbExternRef(Value);
```
saulecabrera, line 75: Is there any advantage on prefixing this with `Rb`? `ExternRefValue(Value)` maybe, since `Value` is what we really work with? I don't think we have adopted the `Rb*` naming elsewhere, aside from whatever we consume from Magnus (if anything).

saulecabrera (approved):

ext/src/ruby_api/store.rs
```diff
@@ -54,6 +55,15 @@ impl StoreData {
...
         self.wasi.as_mut().expect("Store must have a WASI context")
     }
+
+    pub fn retain(&mut self, value: Value) {
+        self.refs.push(value);
+    }
+
+    pub fn mark(&self) {
+        gc::mark(&self.user_data);
+        self.refs.iter().for_each(gc::mark);
```
ianks, line 65: may be able to use `gc::mark_slice(&self.refs)` here to avoid multiple calls into libruby`

ext/src/ruby_api/table.rs
```diff
@@ -0,0 +1,166 @@
...
+};
+use wasmtime::Table as TableImpl;
+
+/// @yard
+/// @rename Wasmtime::Table
+/// Represents a WebAssembly table.
+/// @see https://docs.rs/wasmtime/latest/wasmtime/struct.Table.html Wasmtime's Rust doc
+#[derive(Debug)]
+pub struct Table<'a> {
+    store: StoreContextValue<'a>,
```
ianks, line 20: These lifetimes seem off too me... can table outlive a store?

ext/src/ruby_api/convert.rs
```diff
@@ -84,6 +100,60 @@ impl ToExtern for Value {
...
+            if let Ok(true) = symbol.equal(Symbol::from(*FUNCREF)) {
+                return Ok(ValType::FuncRef);
+            }
+            if let Ok(true) = symbol.equal(Symbol::from(*EXTERNREF)) {
+                return Ok(ValType::ExternRef);
+            }
+        }
+
+        err!(
+            "invalid Webassembly type, expected one of [:i32, :i64, :f32, :f64, :v128, :funcref, :externref], got {:}",
```
ianks, line 151:
```suggestion
            "invalid WebAssembly type, expected one of [:i32, :i64, :f32, :f64, :v128, :funcref, :externref], got {:}",
```

spec/unit/table_type_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    it "creates a table type without maximum" do
+      type = TableType.new(:funcref, 1)
+      expect(type.minimum).to eq(1)
+      expect(type.maximum).to be_nil
+    end
+
+    it "raises on invalid Wasm type" do
+      expect { TableType.new(:nope, 1) }
+        .to raise_error(Wasmtime::Error, /invalid Webassembly type/)
```
ianks, line 20:
```suggestion
        .to raise_error(Wasmtime::Error, /invalid WebAssembly type/)
```

ext/src/ruby_api/table.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        let inner = TableImpl::new(store.context_mut(), tabletype.get().clone(), default_val)
+            .map_err(|e| error!("{}", e))?;
+
+        let table = Self {
+            store: s.into(),
+            inner,
+        };
+
+        table.retain_extern_ref(default)?;
```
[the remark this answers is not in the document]
jbourassa: No strong opinion here, but _maybe_ isn't super explicit. What about `retain_non_nil_extern_ref`?

ext/src/ruby_api/convert.rs
jbourassa: You are right that we do not use this naming style elsewhere. Will update.

ext/src/ruby_api/store.rs
jbourassa: TIL! That works, and I've also applied the same thing to the Linker's marking.

ext/src/ruby_api/table.rs
jbourassa: The table marks the store, therefore the table lives at least as long as the store -- right?

Do you have an example where you think that'd be problematic?

ext/src/ruby_api/table.rs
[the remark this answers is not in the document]
saulecabrera: The thing is that the method isn't explicit either 🤷 it's _maybe_ retaining whatever you send in. Really I'd prefer to remove all ambiguity by having this inline if it is not used elsewhere.

ext/src/ruby_api/store.rs
jbourassa: I think [that might be the culprit](https://github.com/ruby/ruby/blob/4491bb740a9506d76391ac44bb2fe6e483fec952/gc.c#L6185):

```c
static void
gc_mark_locations(rb_objspace_t *objspace, const VALUE *start, const VALUE *end, void (*cb)(rb_objspace_t *, VALUE))
{
    long n;

    if (end <= start) return; // <------ this line
    n = end - start;
    each_location(objspace, start, n, cb);
}
```

`end` is excluded and does not get marked. By reading [Magnus' wrapper](https://github.com/matsadler/magnus/blob/main/src/gc.rs#L37-L53), I understand the intent is to mark the entire slice, so that looks like a bug, right?

```rust
/// Mark multiple Objects.
///
/// Used to mark any stored Ruby objects when implementing
/// [`DataTypeFunctions::mark`](`crate::r_typed_data::DataTypeFunctions::mark`).
pub fn mark_slice<T>(values: &[T])
where
    T: ReprValue,
{
    if let (Some(start), Some(end)) = (values.first(), values.last()) {
        unsafe {
            rb_gc_mark_locations(
                start as *const _ as *const VALUE,
                end as *const _ as *const VALUE,
            )
        }
    };
}
```

This on the other hand works:
```diff
-                end as *const _ as *const VALUE,
+                (end as *const _ as *const VALUE).offset(1),
```

ext/src/ruby_api/table.rs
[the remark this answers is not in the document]
jbourassa: Hmm, with the `Result` it's not as nice as I'd like it to be. I'll just inline for now.

ext/src/ruby_api/store.rs
jbourassa: Opened a PR to Magnus: https://github.com/matsadler/magnus/pull/40

ianks: 🔥

ianks (approved):

PR #72, opened 2022-12-08
Add benchmarks

jbourassa:
Initial benchmark suite. They're not the most helpful benchmarks for now, but just having them there will make it easier to try out optimizations.

Just like examples, I figured we should run them on CI to make sure they're still valid. But... faster, so we don't spend 5+s on each thing we're benchmarking.


jbourassa:
It's indeed fast on CI:
Screenshot 2022-12-08 at 9 35 16 AM

PR #73, opened 2022-12-08
Add global support

### LINKED ISSUE #19: Support Globals
Implement support for Global: https://docs.wasmtime.dev/api/wasmtime/struct.Global.html

jbourassa:
Fixes #19 

Add support for Wasm globals. Decisions worth validating:
1. For creating `GlobalType`s, I've chosen to define 2 constructors (`GlobalType.const` & `GlobalType.var`) instead of accepting mutability as an argument. It feels like a nicer UI, less error prone, and nothing precludes adding `.new` that takes a symbol argument if we ever need that.
2. Following [Wasmtime's naming](https://docs.rs/wasmtime/latest/wasmtime/struct.GlobalType.html#method.content), I'm calling a GlobalType Wasm type "content" (e.g. `GlobalType#content -> :i32`). This naming feels odd to me, but didn't want to diverge without a clear reason.

----

I'm looking into the memcheck failure but it does not look related to this PR, I can reproduce on main (edit: fix for memcheck in #75).

### DIFF
--- a/ext/src/ruby_api/convert.rs
+++ b/ext/src/ruby_api/convert.rs
@@ -2,7 +2,7 @@ use crate::{define_rb_intern, err, error};
 use magnus::{Error, Symbol, TypedData, Value};
 use wasmtime::{ExternRef, Val, ValType};
 
-use super::{func::Func, memory::Memory, store::StoreContextValue, table::Table};
+use super::{func::Func, global::Global, memory::Memory, store::StoreContextValue, table::Table};
 
 define_rb_intern!(
     I32 => "i32",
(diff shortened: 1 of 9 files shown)


saulecabrera (approved):

saulecabrera:
> For creating GlobalTypes, I've chosen to define 2 constructors (GlobalType.const & GlobalType.var instead of accepting mutability as an argument. It feels like a nicer UI, less error prone, and nothing precludes adding .new that takes a symbol argument if we ever need that.

No objections on this. I actually prefer it this way.

PR #76, opened 2022-12-09
Implement Engine epoch timers with Tokio

jbourassa:


ext/src/ruby_api/engine.rs
```diff
@@ -1,15 +1,28 @@
 use super::{config::Config, root};
 use crate::error;
+use lazy_static::lazy_static;
 use magnus::{function, method, scan_args, Error, Module, Object, RString, Value};
+use std::cell::RefCell;
+use tokio::{runtime::Runtime, task::JoinHandle, time};
 use wasmtime::Engine as EngineImpl;
 
+lazy_static! {
+    static ref TOKIO_RT: Runtime = Runtime::new().unwrap();
```
jbourassa, line 10: Note to self: figure out if this can fail in any way.

jbourassa: Doesn't look like anything can actually fail by looking at the code.

jbourassa:
I'm noticing that, on my machine, the tokio-based timer takes about 5% of 1 CPU core, while a similar Ruby script (sleep 1ms, increment epoch, loop) takes 1-2%. That's surprising, and I don't know why yet.

spec/integration/epoch_interruption_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      engine.increment_epoch
+      expect { instance.invoke("42") }.to raise_error(Trap)
+    end
+
+    describe "Engine timer" do
+      it "prevents infinite loop from running forever" do
+        instance = Instance.new(store_deadline_1, mod)
+        engine.start_epoch_timer(10)
+        expect { instance.invoke("loop_forever") }.to raise_error(Trap)
```
jbourassa, line 60: I initially had an assertion here on elapsed time, but it was quite inaccurate in CI, even though it's quite accurate on my M1 mac and on my Linux VM. I've seen [it take less the expected time on Windows](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3662927895/jobs/6192358987), or [14 times as long](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/3664528229/jobs/6195001928).

I don't have a Windows machine to try this out, but I _think_ the clock should still be fairly accurate there. I'm thinking (hoping) it might just be an issue with VMs / CI infrastructure.

ext/src/ruby_api/engine.rs
```diff
@@ -1,15 +1,35 @@
...
+#[magnus::wrap(class = "Wasmtime::Engine", free_immediately)]
 pub struct Engine {
     inner: EngineImpl,
+    timer_task: RefCell<Option<JoinHandle<()>>>,
+}
+
+impl Drop for Engine {
+    fn drop(&mut self) {
+        self.stop_epoch_timer()
+    }
```
jbourassa, line 32: There is no test for that -- couldn't find a sane way of testing: can't read the engine after it's gone.

I did test this manually by adding print statements when ticking, and confirmed that on GC the ticking stopped.

saulecabrera (approved):
LGTM; according to our discussion earlier today, are you planning on adding the `epoch_deadline_callback` in a separate PR?

jbourassa:
> are you planning on adding the `epoch_deadline_callback` in a separate PR?

I wasn't planning on adding it now given we don't have an immediate use case for it and it's not in the other bindings.

Do you think we should add it regardless?

ext/src/ruby_api/engine.rs
```diff
@@ -1,15 +1,35 @@
...
+use std::cell::RefCell;
+use tokio::{runtime, task::JoinHandle, time};
 use wasmtime::Engine as EngineImpl;
 
+lazy_static! {
+    static ref TOKIO_RT: runtime::Runtime = runtime::Builder::new_multi_thread()
+        .thread_name("wasmtime-engine-timers")
+        .worker_threads(1)
+        .enable_time()
+        .enable_io()
```
ianks, line 15: Do we need io?

ext/src/ruby_api/engine.rs
```diff
@@ -26,11 +46,55 @@ impl Engine {
...
+        *self.timer_task.borrow_mut() = Some(handle);
+    }
+
+    /// @yard
+    /// Stops a previously started timer with {#start_epoch_interval}.
+    /// Does nothing if there is no running timer.
+    /// @return [nil]
+    pub fn stop_epoch_interval(&self) {
+        if let Some(handle) = self.timer_task.take() {
+            handle.abort();
```
ianks, line 86: Should we block until the handle `is_cancelled?` I'm wondering what the implications might be otherwise... i.e. if this sequence of events happens, are we safe?

1. Engine is dropped
2. `stop_interval` is called, but there is still a task running
3. Ruby decides to GC the engine
4. Timer calls `increment_epoch`

- I _assume_ we are OK because `Engine` is behind some type of refcell thingy
- But what about our `Store`? Does it being `UnsafeCell` any implications?

spec/integration/epoch_interruption_spec.rb
```diff
@@ -0,0 +1,83 @@
+module Wasmtime
+  RSpec.describe "Epoch interruption" do
```
ianks, line 2: What happens when the trap occurs when we are inside Ruby code? I imagine the VM would not be very happy if we were to interrupt it before it had a chance to make sure it is is a good state.

I'm wondering if we may want to use something like `rb_postponed_job_register_one` to ensure signal safety, and give Ruby a chance to finish up whatever it was trying to do?

ianks: Maybe we want an [InterruptHandle](https://docs.rs/wasmtime/0.19.0/wasmtime/struct.InterruptHandle.html) which does dirty work.

Either way, let’s add some nasty edge case tests for this. Off the top of my head, we need to test:

- calling into a Ruby host method that just `sleep(60)`
- calling into ruby host method that’s generating a ton of garbage in an infinite loop, `with_gc_stress`. Run that about 100 times to make sure we experience an interrupt during GC

saulecabrera:
> I wasn't planning on adding it now given we don't have an immediate use case for it and it's not in the other bindings.

Yep, that's fair. I just wasn't clear if you had the intention to add it right now.

spec/integration/epoch_interruption_spec.rb
saulecabrera: > Hmm, that link is for wasmtime 0.19, looks like it got removed in https://github.com/bytecodealliance/wasmtime/pull/3925. From the changelog:

Indeed, this was removed in favour of epoch based interruption.

saulecabrera: FWIW I don't think that having a test that verifies that function imports are not interrupted makes a lot of sense. We are testing the internals of the engine (Wasmtime) at that point, more than anything else. All the interruption mechanisms(fuel and epoch) are concerned about the execution of the Wasm code, they are not concerned about the function imports provided by the embedding. More concretely, epoch checks are performed on [entering each Wasm function](https://github.com/bytecodealliance/wasmtime/pull/3699/files#diff-f4e65f6eab2ed0d2568b98f7090476dab5cf37bbf7e7d5ef10a6a2ca9e97c33eR635), so if the deadline has passed after executing a function import, the interruption should happen at that point in the program.

ianks (approved):

spec/integration/epoch_interruption_spec.rb
jbourassa: That's a fair point -- we are definitely testing the wasmtime contract here. I'm tempted to leave the test in given it's "cheap" and asserts the behaviour doesn't change on us which would have pretty bad consequences. But if you feel strongly about it I'm happy to remove it too, no strong opinon.

At any rate I'll merge this but it's trivial to remove that commit after, so do not hesitate.

PR #81, opened 2022-12-14
Merge`*Type` on their respective class (`Type`, `Memory`, ...)

jbourassa:
This is a proposal to change the API to address @sandstrom's feedback in #13. The general ideal is to remove the `*Type` (`FuncType`, `MemoryType`, `GlobalType`, `TableType`), collapsing their API surface on the `Func` / `Memory` / etc.

This PR isn't polished (I need to review the docs and proof-read everything), but putting it out there to collect feedback on the idea.

### `Func`

Params and results Array are now on `Func.new`. `FuncType#params` and `FuncType#results` are moved on `Func`.

```ruby
# Before
func_type = FuncType.new([:i32], [:i32])
func = Func.new(store, func_type) { ... }

func_type.params # => [:i32]
func_type.results # => [:i32]

# After
func = Func.new(store, [:i32], [:i32]) { ... }

func.params # => [:i32]
func.results # => [:i32]
```

### `Memory`

Min and max size are now on `Memory.new` as kwargs (`min_size` is required, `max_size` is optional). `MemoryType#minimum` and `#maximum` are now `Memory#min_size` and `#max_size`.

I chose `min` and `max` as it's shorter and is commonly used in Ruby (`Array#min`, `#max`, `#minmax`). 

```ruby
# Before
mem_type = `MemoryType.new(1, 2)`
mem = Memory.new(store, mem_type)

mem_type.minimum # => 1
mem_type.maximum # => 2

# After
mem = Memory.new(store, min_size: 1, max_size: 2)
mem.min_size # => 1
mem.max_size # => 2
```

### `Global`

Moved the `.const` and `.var` constructor from `GlobalType` to `Global`. Moved `GlobalType#content` to `Global#type` -- unifying with `Table` (see below). Unlike Rust, `type` isn't a reserved word in Ruby, and felt more appropriate than `content`.

```ruby
# Before
global_type = GlobalType.var(:i32)
global = Global.new(store, global_type, 1)

global_type.var? # => true
global_type.const? # => false
global_type.element # => :i32

# After
global = Global.var(store, :i32, 1)
global.var? # => true
global.const? # => false
global.type # => :i32
```

### `Table`

Moved type as as positional argument and min_size and max_size as kwarg on `Table.new`.  Moved all `TableType` methods to `Table`, using similar naming as `Memory` (`min_size`, `max_size`) and `Global` (`type`).

```ruby
# Before
table_type = TableType.new(:funcref, 1, 2)
table = Table.new(store, table_type, nil) # nil: default value

table_type.element # => :funcref
table_type.minimum # => 1
table_type.maximum # => 2

# After
table = Table.new(store, :funcref, nil, min_size: 1, max_size: 2)
table.type # => :funcref
table.min_size # => 1
table.max_size # => 2
```


saulecabrera (commented):
I don't have objections on the approach. My main concern (before going over the change) was the overhead that this might introduce, but I'm happy to see that this change is pretty much re-using the validation machinery that was there before (.e.g `FuncType`). Also ruby-wise, this seems more natural.

jbourassa:
@sandstrom Thanks for the feebdback!

> Func
> * I'd use kwargs, easier to extend func constructor with more stuff later, and
> * with kwgargs it'll be more explicit (easier to read/understand the code)

I agree `kwargs` come with an added readability bonus. In a way, that'd match what WAT does (`(func (param i32) (result f64) ...)`). On the other hand, they make everything a little more verbose 🤔 . I'll sleep on this one.

----

> My understanding is that both a _var_ and _const_ would be instances of `Global`.

That's correct. My rationale for that was to avoid the bool flag. I'm tempted to leave this one as is, particularly given it's easy to add `.new` in a non-breaking way later.

>:func_ref would be easier to read in my view.

For that we'll stick with `:funcref` / `:externref` given [that's what's in the spec](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md#language-extensions) and in the WebAssembly Text Format (WAT).

ianks:
As far as design goes, I’m ok with a less vinegary interface given:

1. it doesn’t hurt performance
2. it doesn’t make it harder to auto-generate bindings in the future

There definitely needs to be a more Ruby-ish interface all of this stuff, but the *primary goal of this gem is to have a robust foundational layer for wasmtime in Ruby*.

The ergonomic interfaces will come later in other gems and will be much less of a headache to work with (👋 wit-bindgen-ruby, wasmtime-ffi, who knows).

That all being said, I don’t think this PR goes against points 1 or 2 so I’m conceptual 👍🏻  on it. Just wanted to set clear expectations

jbourassa:
Thanks for chiming in Ian.

> but the _primary goal of this gem is to have a robust foundational layer for wasmtime in Ruby_.

Couldn't agree more. That principle has been guiding the design of the gem. I think it's time I document the project goals in the README.

saulecabrera (approved):

ianks (approved):

PR #84, opened 2022-12-16
More examples

jbourassa:
Add a bunch of examples, essentially copying the wasmtime's examples directory and adapting to Ruby.

I wanted the examples to be dead simple to copy paste. They should have no external dependency (no external helper, no gem, nothing). This means we don't do `assert` (there's no built-in assert in Ruby) and no helper functions.

### DIFF
--- a/examples/linking.rb
+++ b/examples/linking.rb
@@ -1,0 +1,25 @@
+require "wasmtime"
+
+engine = Wasmtime::Engine.new
+
+# Create a linker to link modules together. We want to use WASI with
+# the linker, so we pass in `wasi: true`.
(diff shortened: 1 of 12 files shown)


examples/linking.rb
```diff
@@ -0,0 +1,25 @@
+require "wasmtime"
+
+engine = Wasmtime::Engine.new
+
+# Creat a linker to link modules together. We want to use WASI with
```
saulecabrera, line 5:
```suggestion
# Create a linker to link modules together. We want to use WASI with
```

saulecabrera (approved):

Issue #22, opened 2022-11-14
Missing engine Config


jbourassa:
Not all configuration options are available. Go through the list of Wasmtime's [Config](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html) and implement the ones that are relevant.

jbourassa:
I'm thinking of changing the config to take in a `Hash` of options instead of a `Config` instance, removing the `Config` class altogether. This is quite natural in Ruby and lends itself to this API: the config is basically all scalars, and new engines aren't created _very_ often, so the cost of hashing is negligible.

**Current API:**

```ruby
config = Config.new
config.epoch_interruption = true
config.consume_fuel = true
engine = Engine.new(config)
```

**Proposed API:**

```ruby
engine = Engine.new(epoch_interruption: true, consume_fuel: true)
```


Alternatively, if we think we have to keep the `Config` class, I'd wanna change it to a builder-style API -- I think it's quite elegant for this use case.

```ruby
config = Config.new
  .epoch_interruption(true)
  .consume_fuel(true)
engine = Engine.new(config)
```

PR #89, opened 2022-12-19
Engine config

jbourassa:
Implement more engine configs, closing #22.

The commits are best reviewed individually:

1. Move `static_id` to `helpers` because it's not specific to wasmtime.
2. Add `SymbolEnum`, a struct that represents a `Symbol => T` mapping and use it when converting Symbol to Wasm val types (e.g. `:i32` -> `wasmtime::ValType::I32`).
3. Replace `Wasmtime::Config` with a `Hash` of options on `Wasmtime::Engine.new`, and add more settings.  They're not _all_ supported -- there's a lot and I don't know if they're all relevant. But it should be easy to add new ones as needs arise.


ext/src/helpers/symbol_enum.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Returns an ArgumentError listing all possible valid symbols
+    /// when the passed in `Value` isn't one of the symbol.
+    pub fn get(&self, needle: Value) -> Result<T, magnus::Error> {
+        let needle: Symbol = needle.try_convert().map_err(|_| self.error(needle))?;
+        let id = magnus::value::Id::from(needle);
+
+        self.mapping
+            .0
+            .iter()
+            .find(|(haystack, _)| *haystack == id)
```
jbourassa, line 30: In my experimentation this was faster than a HashMap, which isn't too surprising given our enums are small.

ianks (approved):

PR #92, opened 2022-12-20
Prep v0.4.0 release

jbourassa:
* Bump gem version
* Generate changelog with the `github_changelog_generator` gem


jbourassa:
I pushed this commit directly to the main branch instead.

PR #112, opened 2023-01-17
Implement GC compaction for store data

ianks:
Works with Ruby 2.7+. Can probably implement this for a few more things later.


jbourassa (approved):
That's great! Thanks.

The failures are due to Ruby 2.7 kwargs vs Hash. I don't know how to fix this in a native extension using `rb_scan_args`, and quick Googling didn't help.

It's unfortunate but given 2.7 will stop being supported officially in ~2 months, maybe this could be fixed by changing the test?

PR #113, opened 2023-01-17
Add `Memory#read_utf8`

ianks:
For times when you want to read only a valid utf8 string, a bit more convenient than trying to encode using Ruby.


jbourassa (approved):
👌

PR #114, opened 2023-01-18
Add `Memory#slice` for zero-copy data access

ianks:
Currently, reading data from wasm memory always requires `malloc`ing a new buffer. For certain applications, this can become prohibitively expensive. This PR adds a new method (`Memory#slice`), which returns a readonly view of the `Memory` slice. 

The slice can be converted to a [`Fiddle::MemoryView`](https://docs.ruby-lang.org/en/master/memory_view_md.html) or a frozen string. The frozen string essentially re-implements [`Fiddle::MemoryView#to_s`](https://github.com/ruby/fiddle/blob/36b24325754880266bacf49690e09826b30ad30b/ext/fiddle/memory_view.c#L271) which directly uses the memory slice pointer and length to create a `STR_NO_FREE` string, with an associated ivar to prevent GC.

In the implementation, there is a guard that invalidates the slice whenever the memory size changes. In my tests, this check doesn't seem to matter since the slice is addressed in a position-independent way (using offsets). However, I took the conservative approach and included this check anyway since it seems like a sane invariant.

### DIFF
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -128,7 +128,7 @@ impl<'a> Memory<'a> {
     }
 
     /// @yard
-    /// Read +size+ bytes starting at +offset+ into a `[Wasmtime::Memory::Slice`. This
+    /// Read +size+ bytes starting at +offset+ into a {Slice}. This
     /// provides a way to read a slice of memory without copying the data.
     ///
     /// Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -1,3 +1,7 @@
+mod slice;
+
+use self::slice::Slice;
+
 use super::{
     root,
     store::{Store, StoreContextValue},
@@ -7,6 +11,7 @@ use magnus::{
     function, memoize, method, r_string::RString, r_typed_data::DataTypeBuilder, scan_args,
     DataTypeFunctions, Error, Module as _, Object, RClass, TypedData, Value,
 };
+
 use wasmtime::{Extern, Memory as MemoryImpl};
 
 define_rb_intern!(
@@ -123,6 +128,22 @@ impl<'a> Memory<'a> {
     }
 
     /// @yard
+    /// Read +size+ bytes starting at +offset+ into a `[Wasmtime::Memory::Slice`. This
+    /// proivdes a way to read a slice of memory without copying the data.
+    ///
+    /// @def slice(offset, size)
+    /// @param offset [Integer]
+    /// @param size [Integer]
+    /// @return [Wasmtime::Memory::Slice] Slice of the memory.
+    pub fn slice(
+        rb_self: WrappedStruct<Self>,
+        offset: usize,
+        size: usize,
+    ) -> Result<WrappedStruct<Slice<'a>>, Error> {
+        Ok(Slice::new(rb_self, offset..(offset + size))?.into())
+    }
+
+    /// @yard
     /// Write +value+ starting at +offset+.
     ///
     /// @def write(offset, value)
@@ -159,6 +180,10 @@ impl<'a> Memory<'a> {
     pub fn get(&self) -> MemoryImpl {
         self.inner
     }
+
+    fn data(&self) -> Result<&[u8], Error> {
+        Ok(self.inner.data(self.store.context()?))
+    }
 }
 
 impl From<&Memory<'_>> for Extern {
@@ -168,7 +193,7 @@ impl From<&Memory<'_>> for Extern {
 }
 
 pub fn init() -> Result<(), Error> {
-    let class = root().define_class("Memory", Default::default())?;
+    let class = Memory::class();
     class.define_singleton_method("new", function!(Memory::new, -1))?;
     class.define_method("min_size", method!(Memory::min_size, 0))?;
     class.define_method("max_size", method!(Memory::max_size, 0))?;
@@ -177,6 +202,9 @@ pub fn init() -> Result<(), Error> {
     class.define_method("write", method!(Memory::write, 2))?;
     class.define_method("grow", method!(Memory::grow, 1))?;
     class.define_method("size", method!(Memory::size, 0))?;
+    class.define_method("slice", method!(Memory::slice, 2))?;
+
+    slice::init()?;
 
     Ok(())
 }
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -1,6 +1,6 @@
-mod slice;
+mod unsafe_slice;
 
-use self::slice::Slice;
+use self::unsafe_slice::UnsafeSlice;
 
 use super::{
     root,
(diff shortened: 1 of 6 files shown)


ext/src/ruby_api/memory.rs
```diff
@@ -122,6 +127,22 @@ impl<'a> Memory<'a> {
             .map(RString::new)
     }
 
+    /// @yard
+    /// Read +size+ bytes starting at +offset+ into a `[Wasmtime::Memory::Slice`. This
```
jbourassa, line 131:
```suggestion
    /// Read +size+ bytes starting at +offset+ into a {Slice}. This
```

This will make YARD link to the `Slice` page, and also seems required for the `Slice` class to show up in the sidebar (likely due to bug/missing feature in yard rustdoc which I should fix!).

ext/src/ruby_api/memory/slice.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Gets the memory slice as a Ruby string without copying the underlying buffer.
+    ///
+    /// @def to_str
+    /// @return [String] Binary +String+ of the memory.
+    pub fn to_str(rb_self: WrappedStruct<Self>) -> Result<Value, Error> {
+        let raw_slice = rb_self.get()?.get_raw_slice()?;
+        let id: Id = (*IVAR_NAME).into();
+        let rstring = unsafe {
+            let val = rb_str_new_static(raw_slice.as_ptr() as _, raw_slice.len() as _);
+            rb_ivar_set(val, id.as_raw(), rb_self.as_raw());
```
ianks, line 93: Re-implementation of [`Fiddle::MemoryView#to_s`](https://github.com/ruby/fiddle/blob/36b24325754880266bacf49690e09826b30ad30b/ext/fiddle/memory_view.c#L271)

ianks:
I added some more documentation about the safety and expectations of using this method in https://github.com/bytecodealliance/wasmtime-rb/pull/114/commits/06a2d6abee9423ed735aded66b3b40b80c7e2575

@jbourassa thoughts?

ext/src/ruby_api/memory/unsafe_slice.rs
```diff
@@ -17,33 +17,37 @@ use rb_sys::{
...
 
 /// @yard
-/// @rename Wasmtime::Memory::Slice
+/// @rename Wasmtime::Memory::UnsafeSlice
 /// Represents a slice of a WebAssembly memory. This is useful for creating Ruby
 /// strings from WASM memory without any extra memory allocations.
+///
+/// The returned {Slice} lazily reads the underlying memory, meaning that
+/// the actual pointer to the string buffer is not materialzed until
+/// `Slice#to_str` is called.k
```
jbourassa, line 26:
```suggestion
/// {UnsafeSlice#to_str} is called.
```

ext/src/ruby_api/memory.rs
```diff
@@ -131,19 +131,32 @@ impl<'a> Memory<'a> {
...
     /// provides a way to read a slice of memory without copying the data.
     ///
-    /// Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
+    /// The returned {Slice} lazily reads the underlying memory, meaning that
+    /// the actual pointer to the string buffer is not materialzed until
+    /// `Slice#to_str` is called.
+    ///
+    /// SAFETY: Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
     /// the slice, causing any attempt to read the slice to raise an error.
+    /// However, it is not possible to invalidate the Ruby {String} object after
```
jbourassa:
```suggestion
    /// However, it is not possible to invalidate the Ruby +String+ object after
```

`{}`-style is for linking to types / methods that Yard documents as part of the package -- or at least that's my understanding from using it.

ext/src/ruby_api/memory.rs
```diff
@@ -131,19 +131,32 @@ impl<'a> Memory<'a> {
...
+    /// calling `Slice#to_str`.
```
jbourassa:
```suggestion
    /// calling {Memory::UnsafeSlice#to_str}.
```

ext/src/ruby_api/memory.rs
```diff
@@ -131,19 +131,32 @@ impl<'a> Memory<'a> {
     /// Read +size+ bytes starting at +offset+ into a {Slice}. This
     /// provides a way to read a slice of memory without copying the data.
     ///
-    /// Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
+    /// The returned {Slice} lazily reads the underlying memory, meaning that
+    /// the actual pointer to the string buffer is not materialzed until
+    /// `Slice#to_str` is called.
+    ///
+    /// SAFETY: Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
```
jbourassa:
```suggestion
    /// SAFETY: Resizing the memory (as with {Wasmtime::Memory#grow}) will invalidate
```

ext/src/ruby_api/memory.rs
```diff
@@ -131,19 +131,32 @@ impl<'a> Memory<'a> {
...
     /// the slice, causing any attempt to read the slice to raise an error.
+    /// However, it is not possible to invalidate the Ruby {String} object after
+    /// calling `Slice#to_str`.
+    ///
+    /// As such, the caller must ensure that the Wasmtime {Memory} is not
+    /// resized will the while holding the Ruby string.  Failing to do so could
+    /// result in reading invalid memory.
+    ///
+    /// In general, you should prefer using `Memory#read` or `Memory#read_utf8`
```
jbourassa, line 147:
```suggestion
    /// In general, you should prefer using {Memory#read} or {Memory#read_utf8}
```

ext/src/ruby_api/memory.rs
```diff
@@ -131,19 +131,32 @@ impl<'a> Memory<'a> {
...
+    /// the actual pointer to the string buffer is not materialzed until
+    /// `Slice#to_str` is called.
+    ///
+    /// SAFETY: Resizing the memory (as with `Wasmtime::Memory#grow`) will invalidate
     /// the slice, causing any attempt to read the slice to raise an error.
+    /// However, it is not possible to invalidate the Ruby {String} object after
+    /// calling `Slice#to_str`.
+    ///
+    /// As such, the caller must ensure that the Wasmtime {Memory} is not
+    /// resized will the while holding the Ruby string.  Failing to do so could
```
jbourassa:
```suggestion
    /// resized while holding the Ruby string.  Failing to do so could
```

I think this is what you meant?

jbourassa (approved):
This looks _great_, thanks. Left some documentation nits, but I think you won't be able to refer to `{Memory::UnsafeSlice}` until I fix something in yard-rustdoc (no idea why `{Slice}` worked before?).

Feel free to replace the failing references with `+` and I can swap them back once I have a fix for yard rustdoc.

You should be able to reproduce the doc errors locally with `rake doc:rustdoc doc:yard`.

ext/src/ruby_api/memory.rs
ianks: Ahhh, that makes total sense. Fixing now

PR #121, opened 2023-01-24
Patches for easier usage in a crate context

ianks:
This PR represents the work @jbourassa and I paired on to make `wasmtime-rb` easier to use as a _crate_.

The only difference from our pairing session is that I decided to move the error definition into Rust, since it seems like the most tidy way to do things (i.e. not evaling the file as a string...). The only issue is making `yardoc-rust` understand how to parse the docs when using this pattern. Thoughts @jbourassa?


ext/src/ruby_api/mod.rs
```diff
@@ -4,23 +4,23 @@
...
+pub mod global;
+pub mod instance;
+pub mod linker;
+pub mod memory;
+pub mod module;
+pub mod params;
+pub mod store;
+pub mod table;
+pub mod trap;
+pub mod wasi_ctx_builder;
```
jbourassa, line 23: Since we no longer need to `init` individually, do we still need that change?

jbourassa (commented):
> The only issue is making `yardoc-rust` understand how to parse the docs when using this pattern.

That's not supported indeed. We could introduce a `@class` annotation to define a class. e.g. `@class Error < StandardError`. Because these classes aren't tied to a struct, we'd have have to read rustdocs off functions, and allow multiple `@class` declaration for good measure. Not a big deal.

The part that's annoying is the `Trap` constants; we lose the [nice generated docs](https://bytecodealliance.github.io/wasmtime-rb/main/Wasmtime/Trap.html) for it. I don't know how to fix that.

---

The eval part didn't bother me too much, personally. It's kinda like a require, but from Rust. Besides the fact that it's eval, are there other things about it you don't like?

ianks:
> The eval part didn't bother me too much, personally. It's kinda like a require, but from Rust. Besides the fact that it's eval, are there other things about it you don't like?

What what happen if someone were to `require "wasmtime/error"` elsewhere? I believe that would end up redefining the constants, which could cause problems.

> We could introduce a @class annotation to define a class. e.g. @class Error < StandardError. Because these classes aren't tied to a struct, we'd have have to read rustdocs off functions, and allow multiple @class declaration for good measure. Not a big deal.

Might be a worthwhile feature. I find myself using this memoization pattern a fair bit so it would be nice to support it!

ianks:
@jbourassa brought back in the error.rb bundling, with some more checks / robustness. Also had to make some slight adjustments to `crate-type` so things work properly. Should be good to go!

ext/src/ruby_api/mod.rs
ianks: Fixed

jbourassa (approved):
I have a slight preference for keeping the eval only for users of the crate, so that you only get the downsides of eval (loss of introspection) if you're using the crate directly. But it doesn't bother me too much, and it's also easy to change later. Feel free to address it or not.

ianks:
> I have a slight preference for keeping the eval only for users of the crate, so that you only get the downsides of eval (loss of introspection) if you're using the crate directly. But it doesn't bother me too much, and it's also easy to change later. Feel free to address it or not.

For now I'll ship as is. But something to circle back on if that `(eval)` thing becomes annoying (at least more annoying than multiple load behaviors 😬).

PR #124, opened 2023-01-30
Fix linter

jbourassa:
I don't know what changed but the linter started complaining about  `{}` placeholders in `format!`. That's why CI is failing on main.

This PR fixes that.


saulecabrera:
It might be related to Rust 1.67 (assuming we're building against stable). This happened in other repos too.

saulecabrera (approved):

PR #125, opened 2023-01-31
Add support for profiling configuration

dylanahsmith:
Add support for passing along the configuration option for JIT profiling of WASM code.


ianks (approved):

ianks:
Not sure why CI isn't running, but this LGTM.

PR #126, opened 2023-01-31
Fix small lint

ianks:
Small lint fix from #125, since CI didn't run


jbourassa (approved):
Oh yeah, let's fix this.

PR #129, opened 2023-02-01
Ensure source gem builds properly for `cdylib`

ianks:
#121 unintentionally broke source gem compilation due to changes with how `crate-type` was specified. This PR fixes it with an upgrade `rb-sys v0.9.59`.

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -4,7 +4,7 @@ name: Build gems
 on:
   workflow_dispatch:
   push:
-    branches: ["main", "cross-gem/*"]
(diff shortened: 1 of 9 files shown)


jbourassa (approved):
Thanks for fixing this.

Looks like CI isn't running, not sure why? It's set to run on all branches:

https://github.com/bytecodealliance/wasmtime-rb/blob/450c9f2c0c12ee1a074500110b73801031214822/.github/workflows/ci.yml#L2-L6

Anyhow I ran the linter locally just in case and it's fine. Main should be green now 🎉.

PR #130, opened 2023-02-02
Make ci.yml workflow more trigger happy

ianks:
Maybe this will work?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,6 @@ name: CI
 
 on:
   push:
-    branches: ["*"]
     tags-ignore: ["v*"]  # Skip CI for releases
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,10 @@
 name: CI
 
 on:
+  workflow_dispatch:
```
jbourassa, line 5: Why not, if that happens in the future at least we'll be able to trigger a run manually 👍

.github/workflows/ci.yml
```diff
@@ -2,9 +2,10 @@
...
   push:
-    branches: ["*"]
-    tags-ignore: ["v*"]  # Skip CI for releases
+    branches: ["**"]
```
jbourassa, line 7: Oh, so things like `a/b` wouldn't match `*` but would match `**`?

jbourassa (approved):

.github/workflows/ci.yml
ianks: I have no idea why it works, to be 100% frank. But hey, YOLO?

PR #132, opened 2023-02-06
Update `rb-sys` to use new `RbSys::ExtensionTask`

ianks:
Simplifies things a decent bit, and allows for a shared `./target` dir for caching. [Read more here](https://github.com/oxidize-rb/rb-sys/pull/144).


jbourassa (approved):
Thanks!

PR #138, opened 2023-02-11
Upgrade Magnus to 0.5.0

matsadler:
Hey, I've just released a [new version of Magnus](https://github.com/matsadler/magnus/releases/tag/0.5.0) and while I was preparing the release I tested it against wasmtime-rb to make sure I hadn't broken anything.

There were a few changes I needed to make to get it compiling and passing the tests, so I figured I'd submit the changes so you don't have to do the same work all over again.

Here's a summary of what changed:

* replace `WrappedStruct` with `magnus::typed_data::Obj`
  - `Obj::get` is infallible
  - doesn't implement `::mark`, but derefs to Value so can be passed to `gc::mark`
* can't return `Vec` of Ruby values, return an `RArray`
* `Into<Value>`/`Into<Id>` replaced with `IntoValue`/`IntoId`
* `Into<RString>` replaced with `RString::new`
* `RArray::try_convert` no longer wraps a single value in an array, use `RArray::to_ary`


jbourassa (commented):
>so I figured I'd submit the changes so you don't have to do the same work all over again.

❤️ thank you so much, that is great. I'll review more thoroughly soon.

jbourassa (approved):

PR #150, opened 2023-02-23
Reduce dependabot frequency

sandstrom:
May I suggest running dependabot less frequently. It's tricky to follow progress in a repo running dependabot 52 times a year. The commit logs are have bad signal/noise ratio.

You'll also spend only 1/4 of the time merging them.

Monthly is still frequent enough to gain all the benefits.


jbourassa (approved):
That's reasonable. Thank you for the suggestion.

PR #151, opened 2023-02-24
Fix rb-sys build warning

jbourassa:
The gemspec's file list currently include all files, including native
extensions. We thus get the following warning:

>⚠️ gemspec includes native artifact (lib/wasmtime/ext.bundle), please remove it.
>⚠️ gemspec includes native artifact (lib/wasmtime/wasmtime_rb.bundle), please remove it.

This PR addresses that.


ianks:
My warning paid off! 😀

ianks (approved):

PR #155, opened 2023-03-09
Bump Wasmtime to 6.0.1

jbourassa:
Includes a security fix: https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-ff4p-7xrq-q5r8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,18 +260,18 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.93.0"
+version = "0.93.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91b18cf92869a6ae85cde3af4bc4beb6154efa8adef03b18db2ad413d5bce3a2"
+checksum = "a7379abaacee0f14abf3204a7606118f0465785252169d186337bcb75030815a"
 dependencies = [
  "cranelift-entity",
 ]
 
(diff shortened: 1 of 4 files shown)


saulecabrera (approved):
LGTM! CI is failing due to a breaking change it seems?

jbourassa:
> LGTM! CI is failing due to a breaking change it seems?

Looks like there's a memory corruption bug, and our test suite only catches it on macOS. I don't know why it's only surfaced now, I can reproduce on main with a specific script.

I tried figuring out the bug by looking at the code and commenting parts of it, but no luck so far. I'll run it in valgrind and see what I find.

The test case that reproduces it reliably on my machine is:

```ruby
fit "crashes" do
  store = Store.new(engine, Object.new)
  func = Func.new(store, [], [:i32, :i32]) { [1, nil] }

  800.times do
    func.call
  rescue => _e
  end
end
```

and run with `rake compile:release; GC_STRESS=1 rake spec`

jbourassa:
I created #156 to track the macOS segfault. Because the segfault bug already exists on main (even though CI didn't catch it), I think it's best to merge this PR even with the failing checks.

saulecabrera:
> I think it's best to merge this PR even with the failing checks. The CVE has a score of 9.9/10, better get that fixed ASAP.

Yeah, makes sense to me. Let's merge this and come back to the segfault after.

Issue #156, opened 2023-03-12
Segfault on macOS in Func error handling


jbourassa:
Func error handling can segfault on macOS. Reproduction cases are in the [`macos-segfault` branch](https://github.com/bytecodealliance/wasmtime-rb/compare/macos-segfault).

[Minmal repro script](https://github.com/bytecodealliance/wasmtime-rb/blob/a7af5dbc72ee2d8d3238d6be6e3f9297d9e7be86/examples/crash.rb): 

```ruby
require "wasmtime"

GC.stress = true
store = Wasmtime::Store.new(Wasmtime::Engine.new, Object.new)
func = Wasmtime::Func.new(store, [], [:i32, :i32]) { [1, nil] }

1_000.times do
  func.call
rescue Wasmtime::ResultError
end
```

To repro, checkout the `macos-segfault` branch, compile the gem with `rake compile:release`, and run `ruby -Ilib examples/crash.rb` or `rake spec`.

Can't reproduce on Linux. Valgrind doesn't see an issue either:

```
$ valgrind -- ruby -Ilib examples/crash.rb
==106779== Memcheck, a memory error detector
==106779== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==106779== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==106779== Command: /home/jbourassa.linux/.rbenv/shims/ruby -Ilib examples/crash.rb
==106779==
Using wasmtime-rb 6.0.1: /home/jbourassa.linux/wasmtime-rb/lib/wasmtime/wasmtime_rb.so
```
Changing the memory allocator (`MallocNanoZone` env var) gives different crash reports when running the test. The minimal repro case doesn't fail with `MallocNanoZone=0` (`MallocNanoZone=0 ruby -Ilib examples/crash.rb`) no idea why.

ianks:
Seems this issue doesn't exist on 5.0

```sh
❯ ruby -Ilib examples/crash.rb
Using wasmtime-rb 5.0.0: /Users/ianks/.gem/ruby/3.2.1/gems/wasmtime-5.0.0-arm64-darwin/lib/wasmtime/3.2/wasmtime_rb.bundle
Done!
❯ ruby --version
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [arm64-darwin22]
❯ uname -a
[...]

PR #157, opened 2023-03-23
Add support for configuring the engine's `target`

ianks:
This will all users to cross-compile modules for various architectures. On top of that, it provides a mechanism to use a generic CPU architecture that doesn't make use of the native CPU features (i.e. if your servers use a mix of old and new `x86_64` chips).

```ruby
engine = Wasmtime::Engine.new(target: "x86_64-unknown-linux-gnu")
precompiled_bytes = engine.precompile_module("(module)")
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
   push:
     branches: ["**"]
-    tags-ignore: ["v**"]  # Skip CI for releases
+    tags-ignore: ["v**"] # Skip CI for releases
   pull_request:
 
 concurrency:
@@ -56,6 +56,7 @@ jobs:
 
       - name: Run ruby tests (hard-mode with GC.stress)
         run: bundle exec rake spec
+        if: matrix.os != 'macos-latest' # these specs are aborting on CI with no logs, maybe OOM killer?
         env:
           GC_STRESS: "true"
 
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -56,6 +56,7 @@ jobs:
 
       - name: Run ruby tests (hard-mode with GC.stress)
         run: bundle exec rake spec
+        if: matrix.os != 'macos-latest' # these specs are aborting on CI with no logs, maybe OOM killer?
```
jbourassa, line 59: #156 is why it's aborting on macOS. Skipping meanwhile is fine by me, but I really want to fix that.

spec/spec_helper.rb
```diff
@@ -68,4 +68,4 @@ def wasi_debug
   end
 end
 
-at_exit { GC.start(full_mark: true) }
+at_exit { GC.start(full_mark: true) } if ENV["GC_AT_EXIT"] == "1"
```
jbourassa, line 71: Can you expand on your rationale behind this change?

jbourassa (approved):
LGTM, thanks!

spec/spec_helper.rb
ianks: It's really only necessary for valgrind, since Ruby might not free everything at exit otherwise from what I gather. Stole it from nokogiri.

.github/workflows/ci.yml
ianks: Good call. We'll re-enable after the fix is in.

PR #158, opened 2023-03-24
Make `Param` be `Copy` and mark exceptions raised from Ruby

### LINKED ISSUE #156: Segfault on macOS in Func error handling

ianks:
~Previously, `Param` held a shared reference to the `Value` it wrapped. The main reason for this was because `wasmtime::ValType` did not implement `Copy`, although it easy can because it's small enum type.~

~The issue with with holding a `&Value` comes when moving the `Param` into the `Vec<Param>` needed to invoke a `Func`. Ruby values must live either on the machine stack, CPU registers, or marked on the heap. Otherwise, Ruby is free the GC these values.~

~However, the compiler knows nothing about GC, so it is free to go buck-wild and entirely optimize away the referenced `Value` off the stack, writing the `Value` directly from CPU registers to the heap. Since the `Value` is now on the heap with nothing to `gc::mark` it, it is free to be collected by the GC.~

~By making `Param` implement `Copy`, we can now safely move the `Value` into the `Vec<Param>` without worrying about the GC because the `Value` is now on the stack, rather than just the reference.~

Welp, my initial hypothesis was wrong (although it seemed to work locally at first). The real issue seems to be that when returning a `magnus::Error(Error(ExceptionClass, Cow::Owned))` from a linked function, the exception can be GC'd before being returned to Ruby. Seems counter intuitive, and I haven't fully diagnosed exactly how it happens... maybe @matsadler can take a look to see if there's a soundness issue somewhere.

Anyway, retaining and marking any `magnus::Error` raised inside of a func seems to resolve the issue. Admittedly, it dirties up the code a bit but still worth it despite the seemingly low probability of the bug occurring with conservative GC.

With respect to the `impl Copy for Param`, I still think it's worthwhile change so I opted to keep it in the PR. I can rip it out into another PR if you'd like.

resolves #156


ianks:
Welp

matsadler:
I can reproduce this (M2 Mac, macOS 13.2.1, Xcode/Command Line Tools 14.2, Rust 1.68.1, Ruby 3.2.0) with the example in #156. It only reproduces with `GC.stress = true`. I get the error:

```sh
ruby(51193,0x1df220140) malloc: Heap corruption detected, free list is damaged at 0x6000005a8590
*** Incorrect guard value: 0
ruby(51193,0x1df220140) malloc: *** set a breakpoint in malloc_error_break to debug
fish: Job 1, 'script.rb' terminated by signal SIGABRT (Abort)
```

(the last line is from my shell, so you might not see if it you're not using `fish`)

If I comment out `lto = "thin"` in `Cargo.toml` it no longer reproduces. With the change in this PR to mark the error (6eccbb9e935e4f6149bcc2962718209aa6b74e01) it no longer reproduces. However, if I comment out the section of code that does the marking (but leave the rest of the changes in place) it also doesn't reproduce, so I'm not sure the marking is what is actually fixing it.

Putting `magnus::Error` in an `anyhow::Error` is probably not a good idea. `anyhow::Error` is basically a `Box`, and the `Exception` variant of `magnus::Error` contains a Ruby `Value`, so you're storing a Value on the heap, so the GC can't see it.

But, the error in question here is a `magnus::Error::Error(ExceptionClass, String)`, where the exception class has been [marked as a GC root/never be GC'd](https://github.com/bytecodealliance/wasmtime-rb/blob/e07410945d4c8722490103a0e47f4624309870f0/ext/src/ruby_api/errors.rs#L20) (and even if it wasn't the Ruby code isn't reassigning the `ResultError` constant, so it should be rooted there too), so I do not understand what is happening here.

I suspect the crash is due to an overzealous optimisation, and the change here is enough to stop that optimisation. I don't know if the optimisation is a fault in the compiler, or if Magnus is missing something to tell the compiler not to perform that optimisation.

---

I don't know if returning some Rust type as an error with the message as a String and constructing the `magnus::Error` from that in `handle_wasm_error` might be a simpler change.

matsadler:
The crash seems to happen on the first call after the function has errored twice. The errors can be non-consecutive, here's a script that demonstrates that:

``` ruby
require "wasmtime"

GC.stress = true
store = Wasmtime::Store.new(Wasmtime::Engine.new, Object.new)
i = 0
func = Wasmtime::Func.new(store, [], [:i32, :i32]) do
  result = if i == 2 || i == 5
    puts "returning bad result"
    [1, nil]
  else
    puts "returning good result"
    [1, 1]
  end
  i += 1
  result
end

1_000.times do |i|
  puts "starting call #{i}"
  func.call
rescue Wasmtime::ResultError
end
```

outputs:

``` text
starting call 0
returning good result
starting call 1
returning good result
starting call 2
returning bad result
starting call 3
returning good result
starting call 4
returning good result
starting call 5
returning bad result
starting call 6
ruby(58318,0x1df220140) malloc: Heap corruption detected, free list is damaged at 0x6000037b8be0
*** Incorrect guard value: 0
ruby(58318,0x1df220140) malloc: *** set a breakpoint in malloc_error_break to debug
fish: Job 1, 'script.rb' terminated by signal SIGABRT (Abort)
```

The `2` and `5` don't seem to be significant, it just needs to error twice.

If I comment out `caller.expire();` in `func::make_func_closure` the crash no longer reproduces.

matsadler:
At the start of the function where the [caller is wrapped in a Ruby object](https://github.com/bytecodealliance/wasmtime-rb/blob/e07410945d4c8722490103a0e47f4624309870f0/ext/src/ruby_api/func.rs#L214) it is giving over ownership of the caller struct to Ruby. My suspicion is the `wrapped_caller` variable is getting optimised away, that value isn't on the stack anymore, and so Ruby thinks it can GC `wrapped_caller`, which drops the caller struct. If this is what's happening, it seems it's only the error branch that gets optimised this way.

Then, later, [the `caller` variable, which is a reference to the wrapped data, is used](https://github.com/bytecodealliance/wasmtime-rb/blob/e07410945d4c8722490103a0e47f4624309870f0/ext/src/ruby_api/func.rs#L270). The `caller.expire()` sets an `Option` to `None`, freeing the data it contained. My guess it this has already been freed when Ruby dropped `wrapped_caller`, then after a couple of double frees the allocator's free list is corrupted.


I don't really know how to prove this. Whatever is happening it seems it's due to an optimisation that's very sensitive to changes. It's very easy to 'fix' the problem with debug logging. Maybe stepping through with a debugger or inspecting the assembly output would confirm it, but I don't have the patience for either.

This theory does suggest that keeping the caller object alive for a little longer would fix the problem, and we can actually do that from the Ruby side (funnily enough by doing the exact thing the `caller.expire()` was meant to guard against):

```
require "wasmtime"

GC.stress = true
store = Wasmtime::Store.new(Wasmtime::Engine.new, Object.new)

c = nil
func = Wasmtime::Func.new(store, [], [:i32, :i32]) do |caller|
  c = caller
  [1, nil]
end

1_000.times do
  func.call
rescue Wasmtime::ResultError
end
```

This does indeed stop the error for me.

I can also stop the problem reproducing by removing the `let caller = wrapped_caller.get();` line, and changing the `caller.expire();` to `wrapped_caller.get().expire();`

---

That said, I do think the change in this PR, to mark the exception contained in `magnus::Error` is also required.

jbourassa (commented):
Ian and Mat: great investigation, thank you so much.

Sorry for being slow to respond, I'm currently out of town. Will look into it once I'm back in a couple days. I'd like to see if I can find the root cause of that double free, but unsure I can get there in a reasonable time.

-----

>That said, I do think the change in this PR, to mark the exception contained in `magnus::Error` is also required.

IIRC I had something similar a while back with `BoxValue`. I removed it to simplify the code (d051229f8f8321e0d10d0516a619b24e5c2dc69c). My thinking was that nothing called into Ruby between the end of the Wasm execution and the time the exception was returned to Ruby, thus no GC could happen. Does that reasoning hold?

ext/src/ruby_api/func.rs
```diff
@@ -179,8 +179,16 @@ impl<'a> Func<'a> {
         let params = Params::new(&func_ty, args)?.to_vec()?;
         let mut results = vec![Val::null(); func_ty.results().len()];
 
-        func.call(context, &params, &mut results)
-            .map_err(|e| store.handle_wasm_error(e))?;
+        match func.call(context, &params, &mut results) {
+            Ok(()) => Ok(()),
+            Err(e) => {
+                if let Ok(Some(err)) = store.take_last_error() {
+                    Err(err)
```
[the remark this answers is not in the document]
ianks: Good call, done in https://github.com/bytecodealliance/wasmtime-rb/pull/158/commits/659c93dbda0ebfc7c23ccdd83b3167187f1f11a1

jbourassa:
>If it turns out to be something that should/can be fixed in Magnus please do let me know.

@matsadler: I don't think that can be fixed in Magnus, I think it's only a gotcha of Ruby's GC 😅. But would loved to be proven wrong. See [investigation notes](https://github.com/bytecodealliance/wasmtime-rb/issues/156#issuecomment-1491890663).

-----

@ianks with regard to this change:
1. Can expand on the benefits of making `Param` `Copy`? I thought the benefit would be a perf win, but the benchmarks I ran looked almost identical, within the margin of error.
2. I _think_ we don't actually need to mark the error given we never allocate new objects between the time we put the error on the heap and hand it back to Ruby. Even the [GC-evil spec](https://github.com/bytecodealliance/wasmtime-rb/pull/158/commits/bcbde09e395709842e4dc6bf3ef52e7632335283) pass without marking the exception (or so in my local test). Am I missing something?

ianks:
> Can expand on the benefits of making Param Copy? I thought the benefit would be a perf win, but the benchmarks I ran looked almost identical, within the margin of error.

It's not a performance thing, it's a soundness thing. Using shared references to Ruby values is very easy to get wrong, since the compiler has more freedom about what it can do with the value. 

Take this example, which has two similar functions -- one passed by `&Value` and the other passes a copied `Value`:


```rust
#[inline(never)]
fn func_with_value_reference(param: &Param) -> rb_sys::VALUE {
    helpers::do_something_in_ruby_vm_that_allocates();

    // the param's VALUE could be GC'd by now
    unsafe { rb_sys::rb_str_cat_cstr(*param.0, helpers::SOME_CSTR) }
}

#[inline(never)]
fn func_with_value_copy(param: &ParamCopy) -> rb_sys::VALUE {
    helpers::do_something_in_ruby_vm_that_allocates();

    // the param's VALUE is guaranteed to be alive on the stack (or in a register)
    unsafe { rb_sys::rb_str_cat_cstr(param.0, helpers::SOME_CSTR) }
}

// Shared reference to Ruby value
struct Param<'a>(&'a rb_sys::VALUE);

// Has a copy of a Ruby value
#[derive(Copy, Clone)]
struct ParamCopy(rb_sys::VALUE);


fn main() {
    let items = helpers::make_vec_of_rstrings();
    let rstring: &rb_sys::VALUE = helpers::get_first_item_in_vec(items.as_slice());

    // sound
    {
        let param_copy = ParamCopy(*rstring); // &Param(rb_sys::VALUE)
        let result = func_with_value_copy(&param_copy);
        unsafe { rb_sys::rb_p(result) }
    }

    // segfaults
    {
        let param_ref = Param(rstring); // &Param(&rb_sys::VALUE)
        let result = func_with_value_reference(&param_ref);
        unsafe { rb_sys::rb_p(result) }
    }
}
```

You can see how similar the two functions look, yet one will blow up. 

```rust
fn func_with_value_copy(param: &ParamCopy) -> rb_sys::VALUE;

fn func_with_value_reference(param: &Param) -> rb_sys::VALUE;
```

IMO, it's best to avoid these footguns altogether by passing `Copy` types. In the case of this PR, the copy is trivial since it is only 2 machine words (aka fat pointer).

> I think we don't actually need to mark the error given we never allocate new objects between the time we put the error on the heap and hand it back to Ruby. Even the [GC-evil spec](https://github.com/bytecodealliance/wasmtime-rb/pull/158/commits/bcbde09e395709842e4dc6bf3ef52e7632335283) pass without marking the exception (or so in my local test). Am I missing something?

In this case, indeed you are correct that the code wont fail _given what we know so far_. But, what happens if we are in a multi-threaded environment and a context switch hits at the wrong time? Or sometime in the future a non-suspecting contributor add's a call to Ruby? Or LLVM introduces a new optimization which re-orders the Ruby calls?

Point being, I think it's better to think about the problem from a semantics / guarantees point of view rather than what the behavior might be today.

As Mat said, `anyhow::Error` is a boxed error, and again we are in the situation where we have an unrooted `Value` on the heap. We know that can cause footguns, IMO let's avoid them altogether if we can.

ianks:
So with the suite of "gnarly" specs I added, Valgrind is tagging one particular one as an "invalid memory read of 4 bytes":

```ruby
it "ensures results are never GC'd" do
  n_times = n_times(max: 100)
  store = Store.new(engine, Object.new)
  big_array = without_gc_stress { Array.new(256) { :i32 } }
  expected_result = without_gc_stress { Array.new(256) { |i| i.to_s.to_i } }

  func = Func.new(store, [], big_array) { Array.new(256) { |i| i } }

  n_times.times do
    expect(func.call).to eq(expected_result)
  end
end
```

I have no clue why, and I cannot figure out exactly where it's happening since we're getting `<unknown stack frame>` right before the call:

```
Invalid write of size 4
  <unknown stack frame>
  <unknown stack frame>
  wasmtime::func::Func::call_unchecked_raw::{{closure}} (func.rs:900)
  wasmtime_runtime::traphandlers::catch_traps::call_closure (traphandlers.rs:227)
```

jbourassa:
> Point being, I think it's better to think about the problem from a semantics / guarantees point of view rather than what the behavior might be today.

That makes sense, I'm convinced. Especially after dealing with that optimization causing the segfault. Thank you for answering my Qs with such depth and patience 🙇.


As for this PR, what about we carve out the 2 code changes out from the insanity spec? That way we could merge the code changes while we keep working on the issue found by valgrind. I don't know if that's helpful but running the test suite with clang's AddressSanitizer also produced errors, maybe it'd provide extra information.

ext/src/ruby_api/params.rs
```diff
@@ -1,30 +1,41 @@
...
-                format!("{} (param index {}) ", exception, self.index),
-            ),
-            _ => error,
-        })
+    fn to_wasmtime_val(self) -> Result<wasmtime::Val, Error> {
+        self.val
+            .to_wasm_val(self.ty.into())
+            .map_err(|error| match error {
+                Error::Error(class, msg) => {
+                    Error::new(class, format!("{} (param index {}) ", msg, self.index))
```
jbourassa, line 31: For symmetry, should we update this message too?
```suggestion
                    Error::new(class, format!("{} (param at index {}) ", msg, self.index))
```

ext/src/ruby_api/params.rs
```diff
@@ -1,30 +1,41 @@
...
+                }
+                Error::Exception(exception) => Error::new(
+                    ExceptionClass::from_value(exception.class().into()).unwrap_or_else(arg_error),
+                    format!("{} (param index {}) ", exception, self.index),
```
jbourassa, line 35: Ditto
```suggestion
                    format!("{} (param at index {}) ", exception, self.index),
```

ext/src/ruby_api/params.rs
```diff
@@ -48,10 +59,53 @@ impl<'a> Params<'a> {
     pub fn to_vec(&self) -> Result<Vec<wasmtime::Val>, Error> {
         let mut vals = Vec::with_capacity(self.0.params().len());
         for (i, (param, value)) in self.0.params().zip(self.1.iter()).enumerate() {
-            let param = Param::new(i, param.clone(), value);
+            let i: u32 = i
+                .try_into()
+                .map_err(|_| Error::new(arg_error(), "too many params"))?;
```
jbourassa, line 64: Oh, if there are more that u32 params?

ext/src/ruby_api/func.rs
```diff
@@ -216,49 +238,59 @@ pub fn make_func_closure(
...
-                            .enumerate()
-                        {
-                            *wasm_val = rb_val.to_wasm_val(&ty).map_err(|e| {
-                                Error::new(
-                                    result_error(),
-                                    format!("{e} (result index {i} in {callable})"),
-                                )
-                            })?;
+        match (callable.call(unsafe { rparams.as_slice() }), results.len()) {
+            // For len=1, accept both `val` and `[val]`
```
jbourassa, line 242: Nit: mind moving the comment so that it's with `RArray::to_ary`?

ext/src/ruby_api/func.rs
```diff
@@ -169,8 +169,10 @@ impl<'a> Func<'a> {
         let params = Params::new(&func_ty, args)?.to_vec()?;
         let mut results = vec![Val::null(); func_ty.results().len()];
 
-        func.call(context, &params, &mut results)
-            .map_err(|e| store.handle_wasm_error(e))?;
+        match func.call(context, &params, &mut results) {
+            Ok(()) => Ok(()),
+            Err(e) => Err(store.handle_wasm_error(e)),
+        }?;
```
jbourassa, line 175: Is there a reason for this change?

.github/workflows/memcheck.yml
```diff
@@ -7,13 +7,18 @@ on:
       ruby-version:
         description: "Ruby version to memcheck"
         required: true
-        default: "3.1"
+        default: "3.2"
```
jbourassa, line 10: 🙈 I missed that one in #168.

spec/integration/crash_spec.rb
jbourassa, line 1: This is no longer relevant with #171, right?

jbourassa (approved):
Left some comments but overall LGTM. Thanks!

spec/integration/epoch_interruption_spec.rb
```diff
@@ -62,7 +62,7 @@ module Wasmtime
         engine.stop_epoch_interval
         store.set_epoch_deadline(1)
 
-        sleep_ms(2)
+        sleep_ms(5)
```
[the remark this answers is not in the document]
ianks: yes i did

ext/src/ruby_api/params.rs
```diff
@@ -48,10 +59,53 @@ impl<'a> Params<'a> {
...
+            let param = Param::new(i, param.clone(), *value);
             vals.push(param.to_wasmtime_val()?);
         }
 
         Ok(vals)
     }
 }
+
+/// A [`wasmtime::ValType`] that is [`Copy`], so it can be stays on the stack
```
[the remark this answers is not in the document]
ianks: Added a comment about that 🙏🏻

PR #159, opened 2023-03-28
Use Magnus macros to implement TypedData trait

matsadler:
This library is a great stress test for Magnus' API, so I've been trying out some Magnus API changes to see how they fair in use.

While I was doing that I noticed there's a lot of implementations of `TypedData` that are near identical to what would be derived.

Magnus doesn't usually allow deriving `TypedData` for types with generics as the derived implementation isn't guaranteed to be correct, and I don't want people trying it, find it compiles, and then hit nasty segfaults.

However, it seems mean to make you write out exactly what the macro would produce, so I [added an `unsafe_generics` attribute][commit] to bypass the error, and I've back ported it to the current release branch of the magnus-macros crate.

The changes here remove the `TypedData` implementations in favour of the macros with the `unsafe_generics` attribute set.

[commit]: https://github.com/matsadler/magnus/commit/90454ff8320ef8d5b03e196a843188e32ce16be2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -830,9 +830,9 @@ dependencies = [
 
 [[package]]
 name = "magnus-macros"
-version = "0.4.0"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85aa71c9891b2732ff1157e1860a1ee578459fd25811fd3d72cc6e32b3fbdfea"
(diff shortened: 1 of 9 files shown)


ianks:
This looks great, thank you @matsadler. I think a `cargo clippy --fix` should make CI pass for ya.

jbourassa (approved):
>This looks great, thank you @matsadler. 

+1, fantastic!

PR #160, opened 2023-03-29
Mark the Gem as Ractor safe

macournoyer:
Just need to call `rb_ext_ractor_safe` in `init`.

An extension is Ractor safe if it has no global variables.

See https://docs.ruby-lang.org/en/master/extension_rdoc.html#label-Appendix+F.+Ractor+support

Note that all objects have to be created inside the Ractor, as the engine, store, etc. are not shareable (yet?). But it works!

### DIFF
--- a/ext/src/lib.rs
+++ b/ext/src/lib.rs
@@ -10,6 +10,7 @@ pub(crate) use ruby_api::*;
 
 #[magnus::init]
 pub fn init() -> Result<(), Error> {
+    #[cfg(ruby_gte_3_0)]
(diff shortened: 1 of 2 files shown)


jbourassa (approved):
Awesome, thanks!

Conceptually I _think_ we could make the `Module` and the `Engine` Ractor-shareable without too much trouble, but I haven't given it much thought.

macournoyer:
`Store` too it seems! Almost got it working... Will send another PR when I do. Thanks Jimmy 🙌

PR #161, opened 2023-03-30
Allowing making Module and Engine Ractor shareable

macournoyer:
Allows:

```ruby
Ractor.make_shareable(engine)
Ractor.make_shareable(mod)
```

So engine and module can be shared between Ractors:

```ruby
r = Ractor.new(engine, mod) do |engine, mod|
  store_data = Object.new
  store = Wasmtime::Store.new(engine, store_data)
  Wasmtime::Instance.new(store, mod).invoke("hello")
end
```

### DIFF
--- a/ext/src/ruby_api/engine.rs
+++ b/ext/src/ruby_api/engine.rs
@@ -1,3 +1,4 @@
+use std::sync::Mutex;
 use super::{config::hash_to_config, root};
 use crate::error;
 use magnus::{function, method, scan_args, Error, Module, Object, RHash, RString, Value};
@@ -21,7 +22,7 @@ pub struct Engine {
     inner: EngineImpl,
 
     #[cfg(feature = "tokio")]
-    timer_task: std::cell::RefCell<Option<tokio::task::JoinHandle<()>>>,
+    timer_task: Mutex<Option<tokio::task::JoinHandle<()>>>,
 }
 
 #[cfg(feature = "tokio")]
@@ -96,7 +97,7 @@ impl Engine {
             }
         });
 
-        *self.timer_task.borrow_mut() = Some(handle);
+        *self.timer_task.lock().unwrap() = Some(handle);
     }
 
     /// @yard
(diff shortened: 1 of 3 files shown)


ext/src/ruby_api/engine.rs
```diff
@@ -16,7 +16,7 @@ lazy_static::lazy_static! {
 /// @yard
 /// Represents a Wasmtime execution engine.
 /// @see https://docs.rs/wasmtime/latest/wasmtime/struct.Engine.html Wasmtime's Rust doc
-#[magnus::wrap(class = "Wasmtime::Engine", free_immediately)]
+#[magnus::wrap(class = "Wasmtime::Engine", free_immediately, frozen_shareable)]
```
jbourassa, line 19: The `timer_task` field uses a `RefCell` which I think isn't thread safe. IIUC, by making this `frozen_shareable`, we're saying that instances are frozen and can be used across threads _at the same time_.

Should we change that `RefCell` for a `Mutex`?

ext/src/ruby_api/module.rs
```diff
@@ -7,7 +7,7 @@ use wasmtime::Module as ModuleImpl;
 /// Represents a WebAssembly module.
 /// @see https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html Wasmtime's Rust doc
 #[derive(Clone)]
-#[magnus::wrap(class = "Wasmtime::Module", size, free_immediatly)]
+#[magnus::wrap(class = "Wasmtime::Module", size, free_immediatly, frozen_shareable)]
```
jbourassa, line 10: 👍, the underlying Wasmtime Module is thread-safe.

ext/src/ruby_api/engine.rs
macournoyer: Damn! Would be nice if Magnus required `Sync` on the type as soon as we use this flag.

Seems it does, but only on the `DataTypeFunctions`: https://github.com/matsadler/magnus/blob/f7aa98abbed4a1d6e2b3aed19c9e1f2064f0350e/src/typed_data.rs#L364-L375.

jbourassa:
Thanks!

PR #165, opened 2023-04-03
chore(ext): Update extension crate authors

saulecabrera:
So that it matches the authors defined in the Gem specification.


jbourassa (approved):
Fine by me, but it's not my name this is removing 😅. Let's see what Ian says.

ianks (approved):
💯

PR #171, opened 2023-04-07
Add "insanity" specs

ianks:
Extracted from #158

### DIFF
--- a/suppressions/ruby-3.1.supp
+++ b/suppressions/ruby-3.1.supp
@@ -1,20 +1,4 @@
 {
-  Rust probes for statx(buf)
-  Memcheck:Param
-  statx(buf)
-  ...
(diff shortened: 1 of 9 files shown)


suppressions/ruby-3.1.supp
```diff
@@ -1,19 +1,3 @@
-{
```
ianks, line 1: I added these upstream (https://github.com/Shopify/ruby_memcheck/pull/15)

jbourassa (approved):

ianks:
@jbourassa I managed to find the threshold at which Valgrind indicates we have an invalid memory write. It happens at exactly when the results length is > 175 items. I don't imagine this is a common result size, so I opted for raising an argument error when we detect > 174 results until we can find the underlying issue.

PR #175, opened 2023-04-22
Remove warning from Ractor spec

jbourassa:
Ractors are experimental, thus Ruby warns when using them. We don't need to be reminded every time we run the tests though, so let's silence the experimental warning when running those tests.

Also remove the check for Ruby version since all versions we support now have Ractors.


saulecabrera (approved):
TIL about `config.around(...)`

PR #177, opened 2023-04-26
Add docs for fork safety

### LINKED ISSUE #174: Add documentation about fork safety with `Wasmtime::Engine`
Most Ruby servers use some type of pre-forking mechanism to provide parallel request processing. Unfortunately, this can causes processes to deadlock if you compile Wasm before forking (when using `Wasmtime::Engine.new(parallel_compilation: true)`).

As such, if any compilations needs to be done before forking, you need to use `Wasmtime::Engine.new(parallel_compilation: false)`. Let's add something like this to the docs.

ianks:
And a helper to make doc editing a bit nicer with `rake doc:serve`

resolves #174

### DIFF
--- a/ext/src/ruby_api/engine.rs
+++ b/ext/src/ruby_api/engine.rs
@@ -18,6 +18,25 @@ lazy_static::lazy_static! {
 
 /// @yard
 /// Represents a Wasmtime execution engine.
+///
+/// @example Disabling parallel compilation
+///    # Many Ruby servers use a pre-forking mechanism to allow parallel request
+///    # processing. Unfortunately, this can causes processes to deadlock if you
+///    # use parallel compilation to compile Wasm prior to calling
+///    # `Process::fork`. To avoid this issue, any compilations that need to be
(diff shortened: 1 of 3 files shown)


ext/src/ruby_api/engine.rs
```diff
@@ -18,6 +18,25 @@ lazy_static::lazy_static! {
 
 /// @yard
 /// Represents a Wasmtime execution engine.
+///
+/// @example Disabling parallel compilation
+///    # Many Ruby servers use a pre-forking mechanism to allow parallel request
+///    # processing. Unfortunately, this can causes processes to deadlock if you
+///    # use parallel compilation to compile Wasm prior to calling
+///    # `Process::fork`. To avoid this issue, any compilations that need to be
```
jbourassa, line 26:
```suggestion
///    # `Process.fork`. To avoid this issue, any compilations that need to be
```

While totally valid, the `::` notation for class method feels less common than the `.` notation. No strong opinion though!

jbourassa (approved):
Thanks Ian!

PR #197, opened 2023-06-29
Pin nightly and use the same magnus version

saulecabrera:
This change includes:

* Pin Nightly to 2023-05-22: To avoid issues with doc generation, we've pinned the nightly version to 2023-05-22.

* Update the Magnus version for Rust Example: I noticed a discrepancy in the Magnus version being used for the Rust example and the extension. This was causing compilation issues. I've updated the Magnus version in the Rust example to match the one used in the extension to resolve these issues.

* Fix clippy warnings: With the release of the newest stable Rust version, some new clippy warnings were raised. I've addressed these warnings in this update.


saulecabrera:
I was able to reproduce the failures with `nightly-05-22+`

jeffcharles (approved):

jbourassa:
Alternatively, using updating proc_macro2 should work: https://github.com/bytecodealliance/wasmtime-rb/compare/jb/update-all-deps

saulecabrera:
> Alternatively, using updating proc_macro2 should work: https://github.com/bytecodealliance/wasmtime-rb/compare/jb/update-all-deps

Oh this makes sense to me. Do we gain anything significant by not pinning nightly? As far as I understand we only rely on nightly for doc generation for which a pinned nightly should be enough? I'd like to avoid PRs randomly breaking due to un-pinned nightly versions. Perhaps updating proc-macro2 _and_ pinning nightly to the latest makes sense? Let me know if I'm missing something here though!

saulecabrera:
After discussing this change offline with @jbourassa and making several iterations, this is now ready for another review. I plan to defer the updates for dependencies to a future PR, given that this current PR already contains a mix of changes. However, these changes are necessary to make the CI pass again.

jbourassa (approved):
LGTM, thanks!

PR #211, opened 2023-08-11
Address Ruby dep issues and bump to `v9.0.4`

ianks:
Due to issues surrounding ruby-head API stability, we've had to use a web of git forks to make things work with `ruby-head`, which we continuously test in production. Now that https://github.com/oxidize-rb/rb-sys/issues/227 is fixed and [magnus v0.5.5](https://github.com/matsadler/magnus/releases/tag/0.5.5) is released, we can untangle that.

This PR does the following:

1. Upgrades magnus and rb-sys for the `stable-api` feature
2. Upgrade wasmtime to v9.0.4
3. Bump the gem the v9.0.4

After this goes through, I will release a new gem version (v9.0.4).


ext/Cargo.toml
```diff
@@ -16,20 +16,31 @@ ruby-api = []
 
 [dependencies]
 lazy_static = "1.4.0"
-magnus = { version = "0.5.3", features = ["rb-sys-interop"] }
-rb-sys = "~0.9.81"
-wasmtime = { version = "= 9.0.1" }
-wasmtime-wasi = "= 9.0.1"
-wasi-common = "= 9.0.1"
-wasi-cap-std-sync = "= 9.0.1"
+# For ruby-head support, we have to lock to 0.5.4 until
```
saulecabrera, line 19: Should this comment be updated to say `0.5.5`?

ext/Cargo.toml
```diff
@@ -16,20 +16,31 @@ ruby-api = []
...
+wasmtime = { version = "= 9.0.4" }
+wasmtime-wasi = "= 9.0.4"
+wasi-common = "= 9.0.4"
+wasi-cap-std-sync = "= 9.0.4"
 cap-std = "1.0.5"
 anyhow = "*" # Use whatever Wasmtime uses
-wat = "1.0.69"
-tokio = { version = "1.28.2", features = ["rt", "rt-multi-thread", "time", "net"], optional = true }
-async-timer = { version = "1.0.0-beta.8", features = ["tokio1"], optional = true }
+wat = "1.0.63"
```
saulecabrera, line 31: Any reason why we went back to `1.0.63`?

ext/Cargo.toml
```diff
@@ -16,20 +16,31 @@ ruby-api = []
...
+tokio = { version = "1.28.1", features = [
```
saulecabrera, line 32: Same here, any reason why we went back to 1.28.1?

saulecabrera (commented):
Generally looks good to me. Left a couple of comments regarding the versions of a couple of dependencies.

lib/wasmtime/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Wasmtime
-  VERSION = "9.0.1"
+  VERSION = "9.0.4.rc1"
```
jbourassa: What's the rationale behind a RC release?

jbourassa: err I see it's now 9.0.4, resolving.

jbourassa:
Thanks Ian 🙇

ext/Cargo.toml
```diff
@@ -16,20 +16,31 @@ ruby-api = []
...
 [dependencies]
 lazy_static = "1.4.0"
-magnus = { version = "0.5.3", features = ["rb-sys-interop"] }
-rb-sys = "~0.9.81"
-wasmtime = { version = "= 9.0.1" }
-wasmtime-wasi = "= 9.0.1"
-wasi-common = "= 9.0.1"
-wasi-cap-std-sync = "= 9.0.1"
+# For ruby-head support, we have to lock to 0.5.4 until
+# rb-sys/stable-api-compiled-fallback feature is used in Magnus.
```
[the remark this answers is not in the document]
ianks: Fixed

ext/Cargo.toml
ianks: Git conflict resolution gone wrong, fixed now

ext/Cargo.toml
ianks: Fixed

jbourassa (approved):

PR #218, opened 2023-08-22
chore: Update to wasmtime@10

saulecabrera:
Also:
* Clones `ValType` where applicable, given that it's no longer `Copy`
* Updates `magnus` to match the version used in `ext/`

[Changes in version 10](https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#1000)


jbourassa (approved):
I guess you could also bump the gem version in the same PR if you'd like.

ext/Cargo.toml
```diff
@@ -20,10 +20,10 @@ magnus = { version = "0.5.5", features = ["rb-sys-interop"] }
 rb-sys = { version = "*", default-features = false, features = [
   "stable-api-compiled-fallback",
 ] }
-wasmtime = { version = "= 9.0.4" }
-wasmtime-wasi = "= 9.0.4"
-wasi-common = "= 9.0.4"
-wasi-cap-std-sync = "= 9.0.4"
+wasmtime = { version = "= 10.0.0" }
```
ianks, line 23: Should we do 10.0.1 (which has a wasi fix)?

ianks (approved):

PR #225, opened 2023-09-26
chore: Update to wasmtime 13.0.0

saulecabrera:
* Updates to `wasmtime` 13.0.0
* Update the usage of the `WasiCtxBuilder` which as of this update its methods return `&mut Self` https://github.com/bytecodealliance/wasmtime/pull/6770

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ianks:
I think https://github.com/oxidize-rb/actions/releases/tag/v1.1.8 should fix the windows builds. I re-ran the build to see.

saulecabrera:
Thanks for looking into this @ianks 🥇, that release indeed fixed the issues with Windows.

jbourassa (approved):

ianks (approved):

PR #235, opened 2023-10-05
chore: Update wat

saulecabrera:
This commit updates `wat` and fixes the wat examples according to https://redirect.github.com/bytecodealliance/wasm-tools/issues/1215

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -44,9 +44,9 @@ jobs:
       - name: Install deps
         run: |
           bundle config unset deployment
-          bundle add ruby_memcheck --version '~> 1.3.1' & # avoid usage in Gemfile bc it pulls in nokogiri
-          sudo apt-get install -y valgrind &
-          wait
+          bundle add ruby_memcheck --version '~> 1.3.1' # avoid usage in Gemfile bc it pulls in nokogiri
+          sudo apt-get update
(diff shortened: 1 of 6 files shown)


.github/workflows/memcheck.yml
```diff
@@ -44,9 +44,9 @@ jobs:
       - name: Install deps
         run: |
           bundle config unset deployment
-          bundle add ruby_memcheck --version '~> 1.3.1' & # avoid usage in Gemfile bc it pulls in nokogiri
-          sudo apt-get install -y valgrind &
-          wait
+          bundle add ruby_memcheck --version '~> 1.3.1' # avoid usage in Gemfile bc it pulls in nokogiri
+          sudo apt-get update
+          sudo apt-get install -y valgrind
```
saulecabrera, line 49: The memcheck job was failing because it couldn't find `valgrind`; I'm almost certain that running `sudo apt-get update` is what's required to properly install valgrind, however, I was not able to make it work only by updating the packages, I also had to change how these commands are executed: essentially running them inline instead of in the background (with `&` + `wait`). Is there a particular reason to run them in the background?

ianks: Mostly because valgrind was very slow to install. But I'm ok to remove that optimization.

ianks (approved):

PR #238, opened 2023-10-27
Add support for using Winch as a compiler strategy

jeffcharles:
Adds support for setting the `strategy` config option in Wasmtime. Also enables the Winch feature in Wasmtime so the `winch` strategy can be used.

### DIFF
--- a/ext/src/ruby_api/config.rs
+++ b/ext/src/ruby_api/config.rs
@@ -107,7 +107,7 @@ pub fn hash_to_config(hash: RHash) -> Result<Config, Error> {
             config.profiler(entry.try_into()?);
         } else if *CRANELIFT_OPT_LEVEL == id {
             config.cranelift_opt_level(entry.try_into()?);
-        } else if *STRATEGY == id {
+        } else if *STRATEGY == id && cfg!(feature = "winch") {
             config.strategy(entry.try_into()?);
         } else if *TARGET == id {
             let target: Option<String> = entry.try_into()?;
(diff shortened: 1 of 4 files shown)


jeffcharles:
I'm seeing:

```
For crate cargo-cache: Failed to download from remote: could not GET
    │ https://crates.io/api/v1/crates/cargo-cache/0.8.3/download: HTTP status
    │ server error (503 Service Unavailable) for url (https://crates.io/api/v1/
    │ crates/cargo-cache/0.8.3/download)
```

appearing in the CI failures. Might make sense to retry them. I can't seem to trigger a retry.

saulecabrera:
I was not able to re-run them either: it seems that the option to re-run only appears when the entire suite has finished, so I suspect that's the same issue you are experiencing.

jeffcharles:
I was able to retry and it looks like the 503s have cleared up

ext/src/ruby_api/config.rs
```diff
@@ -25,13 +25,17 @@ define_rb_intern!(
     WASM_MEMORY64 => "wasm_memory64",
     PROFILER => "profiler",
     CRANELIFT_OPT_LEVEL => "cranelift_opt_level",
+    STRATEGY => "strategy",
```
jeffcharles, line 28: I've opted to not put the parts outside the conditional behind a `cfg` flag because it makes integrating support into the conditional much more straightforward. But if you'd like, I can change the conditional to use an early return or look into refactoring it into a match statement (or another option) instead so it's easier to put more of the plumbing for the strategy option behind the `cfg` flag.

saulecabrera (approved):

PR #247, opened 2023-11-29
Allow `generate_address_map` config in `Engine`

Maaarcocr:
# why?

turning this flag off makes the precompiled modules considerably smaller


ext/src/ruby_api/engine.rs
```diff
@@ -77,6 +77,7 @@ impl Engine {
     /// @option config [Boolean] :wasm_multi_memory
     /// @option config [Boolean] :wasm_memory64
     /// @option config [Boolean] :parallel_compilation (true) Whether compile WASM using multiple threads
+    /// @option config [Boolean] :generate_address_map
```
ianks, line 80:
```suggestion
    /// @option config [Boolean] :generate_address_map Configures whether compiled artifacts will contain information to map native program addresses back to the original wasm module. This configuration option is `true` by default. Disabling this feature can result in considerably smaller serialized modules.
```

PR #255, opened 2023-12-12
Allow perfmap as a profiling config

Maaarcocr:
# why? 

perfmap is an allowed value as a profiling strategy (and it's also the most cross platform, it works on mac too if you use samply!)

# how?

just add the symbol and add it to the mapping


saulecabrera (commented):
Looks good to me; cargo fmt is complaining. Once that's fixed, we should be ready to land this.

saulecabrera (approved):

PR #266, opened 2024-01-03
chore: Update to `wasmtime` 15

saulecabrera:
This patch updates Wasmtime to version 15.

Changelog: https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#1500

### DIFF
--- a/ext/src/ruby_api/caller.rs
+++ b/ext/src/ruby_api/caller.rs
@@ -77,29 +77,22 @@ impl<'a> Caller<'a> {
     }
 
     /// @yard
-    /// (see Store#fuel_consumed)
-    pub fn fuel_consumed(&self) -> Result<Option<u64>, Error> {
-        self.handle.get().map(|c| c.fuel_consumed())
+    /// @see Store#get_fuel
+    pub fn get_fuel(&self) -> Result<u64, Error> {
+        self.handle
+            .get()
+            .map(|c| c.get_fuel())?
+            .map_err(|e| error!("{}", e))
     }
 
     /// @yard
-    /// (see Store#add_fuel)
-    /// @def add_fuel(fuel)
-    pub fn add_fuel(&self, fuel: u64) -> Result<(), Error> {
+    /// @see Store#set_fuel
+    pub fn set_fuel(&self, fuel: u64) -> Result<(), Error> {
         self.handle
             .get_mut()
-            .and_then(|c| c.add_fuel(fuel).map_err(|e| error!("{}", e)))?;
+            .and_then(|c| c.set_fuel(fuel).map_err(|e| error!("{}", e)))?;
 
         Ok(())
-    }
-
-    /// @yard
-    /// (see Store#consume_fuel)
-    /// @def consume_fuel(fuel)
-    pub fn consume_fuel(&self, fuel: u64) -> Result<u64, Error> {
-        self.handle
-            .get_mut()
-            .and_then(|c| c.consume_fuel(fuel).map_err(|e| error!("{}", e)))
     }
 
     pub fn context(&self) -> Result<StoreContext<StoreData>, Error> {
@@ -121,9 +114,8 @@ pub fn init() -> Result<(), Error> {
     let klass = root().define_class("Caller", class::object())?;
     klass.define_method("store_data", method!(Caller::store_data, 0))?;
     klass.define_method("export", method!(Caller::export, 1))?;
-    klass.define_method("fuel_consumed", method!(Caller::fuel_consumed, 0))?;
-    klass.define_method("add_fuel", method!(Caller::add_fuel, 1))?;
-    klass.define_method("consume_fuel", method!(Caller::consume_fuel, 1))?;
+    klass.define_method("get_fuel", method!(Caller::get_fuel, 0))?;
+    klass.define_method("set_fuel", method!(Caller::set_fuel, 1))?;
 
     Ok(())
 }
--- a/ext/src/ruby_api/caller.rs
+++ b/ext/src/ruby_api/caller.rs
@@ -77,7 +77,8 @@ impl<'a> Caller<'a> {
     }
 
     /// @yard
-    /// @see Store#get_fuel
+    /// (see Store#get_fuel)
+    /// @def get_fuel
     pub fn get_fuel(&self) -> Result<u64, Error> {
         self.handle
             .get()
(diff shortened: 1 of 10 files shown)


ext/src/ruby_api/store.rs
```diff
@@ -163,40 +163,28 @@ impl Store {
...
-    /// Returns the amount of fuel consumed by this {Store}’s execution so far,
-    /// or +nil+ when the {Engine}’s config does not have fuel enabled.
-    /// @return [Integer, Nil]
-    pub fn fuel_consumed(&self) -> Option<u64> {
-        self.inner_ref().fuel_consumed()
+    /// Returns the amount of fuel in the {Store}.
+    /// This function errors if fuel consumption is not configured by passing
+    /// enabling fuel consumption via Wasmtime::Engine::new.
+    ///
+    /// @return [Integer, Error]
```
jbourassa, line 170: The exception isn't returned but raised, we should probably leave it as `@return [Integer]`

ext/src/ruby_api/caller.rs
jbourassa, line 1: The doc are a little off now, I think because we dropped the `(see ...)`. Those are [reference tags](https://rubydoc.info/gems/yard/file/docs/GettingStarted.md#reference-tags) which copy the docs from the things that's being referenced.

ext/src/ruby_api/store.rs
```diff
@@ -163,40 +163,28 @@ impl Store {
...
+    pub fn get_fuel(&self) -> Result<u64, Error> {
+        self.inner_ref().get_fuel().map_err(|e| error!("{}", e))
     }
 
     /// @yard
-    /// Adds fuel to the {Store}.
+    /// Sets fuel to the {Store}.
     /// @param fuel [Integer] The fuel to add.
```
jbourassa, line 177: Doc nit: if I understand correctly, this will _set_ the fuel to the new value, not _add_, right?

ext/src/ruby_api/store.rs
```diff
@@ -163,40 +163,28 @@ impl Store {
...
-    /// @def add_fuel(fuel)
-    /// @return [Nil]
-    pub fn add_fuel(&self, fuel: u64) -> Result<(), Error> {
+    /// @def set_fuel(fuel)
+    /// @return [Error]
```
jbourassa, line 179: From Ruby, this won't _return_ an error but raise it. So the return value would always be nil.

jbourassa (commented):
I wonder if we could instead do `#fuel=` and `#fuel` 🤔. That'd feel more natural in Ruby, although we don't have another case of defining a setter in the library AFAIK.

ext/src/ruby_api/store.rs
saulecabrera: I think that what I need is a `@raise`. I want to signal that this method raises. I've added that here and below.

ext/src/ruby_api/store.rs
saulecabrera: Same as my comment above, I've added `@raise`

saulecabrera:
>  That'd feel more natural in Ruby, although we don't have another case of defining a setter in the library AFAIK.

I considered this, but aside from it being the most natural thing to do in Ruby, I didn't find another good reason to deviate from the approach already followed in the codebase, which is (i) an exact or almost-exact mapping to the Rust API and (ii) mostly using `set_*` instead of `=` for setters; a good example is [`set_epoch_deadline`](https://github.com/bytecodealliance/wasmtime-rb/blob/main/spec/integration/epoch_interruption_spec.rb#L63), but there are other instances too, like some of the methods in `WasiCtxBuilder`.  I think it's reasonable to consider doing this, but I'm not sure if this PR is necessarily the best place to do so, at least I'd prefer having a separate discussion on how/why we want to redesign these specific bits of the API.

ext/src/ruby_api/store.rs
```diff
@@ -163,40 +163,29 @@ impl Store {
...
-    /// Returns the amount of fuel consumed by this {Store}’s execution so far,
-    /// or +nil+ when the {Engine}’s config does not have fuel enabled.
-    /// @return [Integer, Nil]
-    pub fn fuel_consumed(&self) -> Option<u64> {
-        self.inner_ref().fuel_consumed()
+    /// Returns the amount of fuel in the {Store}.
+    ///
+    /// @return [Integer]
+    /// @raise [Error] if fuel consumption is not enabled via
+    /// Wasmtime::Engine::new.
```
jbourassa, line 170: Looks like Yard does not pick up `Wasmtime::Engine::new` like this, maybe it needs to be indented or on the same line? Screenshot 2024-01-05 at 1 07 33 PM

This seems to do it for me:

```suggestion
    /// @raise [Error] if fuel consumption is not enabled via
    ///   {Wasmtime::Engine.new}.
```

- using `{}` will link to the symbol
- while `::` is valid for singleton method call, Yard does not undersand it and I think `.` is more common in the community

Same below.

ext/src/ruby_api/store.rs
jbourassa: Oh, that's nice 👍. We don't apply that consistently in the library but we should!

jbourassa (approved):
>I think it's reasonable to consider doing this, but I'm not sure if this PR is necessarily the best place to do so

That's a reasonable counter-argument. FWIW `set_epoch_deadline` does not have a matching getter, maybe that's why I didn't notice it before -- `get_*` does stand out more to me. At any rate, we can discuss this elsewhere, and it doesn't have to be a breaking change: we can support both.

----

I left one more doc related nit, but looks good. Thank you.

ianks (approved):
Awesome, before releasing let me get 3.3 precompiled binaries configured

sandstrom:
> I think it's reasonable to consider doing this, but I'm not sure if this PR is necessarily the best place to do so, at least I'd prefer having a separate discussion on how/why we want to redesign these specific bits of the API.

Just to chime in a bit on this, since I've also mentioned it previously.

I totally see the point of staying fairly close to the Rust API! But here I'd definitely vote for going with the Ruby convention.

Getters/setters is a common concept, so deciding that any getters/setters would be implemented with Ruby principles in this gem/library, instead of using `get_` and `set_` prefixes, would easy to explain/follow and wouldn't take this library far away from the Rust APIs that it is based on, i.e. the methods would be `fuel` and `fuel=` instead of `get_fuel` [doc](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html#method.get_fuel) and `set_fuel` [doc](https://docs.wasmtime.dev/api/wasmtime/struct.Store.html#method.set_fuel).

(obviously outside the scope of this PR, but I think it would make the library a little less confusing for people that aren't aware of the Rust APIs that it's based on)

PR #270, opened 2024-01-05
Add support for precompiled binaries on Ruby 3.3

ianks:
This mostly just amounts to an rb-sys upgrade, but I moved a couple of steps around in CI for easier debugging/testing


.github/workflows/memcheck.yml
```diff
@@ -7,11 +7,11 @@ on:
       ruby-version:
         description: "Ruby version to memcheck"
         required: true
-        default: "3.2"
+        default: "3.3"
         type: choice
         options:
           - "head"
-          - "3.2"
+          - "3.3"
```
jbourassa, line 14: Let's keep 3.2 here?

jbourassa (approved):

jbourassa:
Thanks! It's been on my mind too.

jbourassa:
BTW CI was failing on main due to linting errors, so we'll likely have to rebase on top of #266 (or main once that PR merges).

PR #276, opened 2024-01-09
[WIP] Deterministic Wasi Context Option

cameronbarker:
To increase security, we will add wasi-common's deterministic [clock](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/src/clocks.rs) and [random](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-common/src/random.rs) configurability to the ruby library.

Approach:

1. Create a new wasi context using the [Shopify/deterministic-wasi-ctx](https://github.com/Shopify/deterministic-wasi-ctx).
2. Update Wasmtime::Store to take either `WasiCtxBuilder` or `WasiDeterministicCtxBuilder`. 
**Note**: The current implementation has introduced a new kwarg `wasi_det_ctx`. I believe there is a better approach and will apply it during a refactor, after confirming this is working with our project.


ext/src/ruby_api/store.rs
```diff
@@ -115,19 +117,25 @@ impl Store {
...
         let user_data = user_data.unwrap_or_else(|| ().into_value());
-        let wasi = match kw.optional.0 {
-            None => None,
-            Some(wasi_ctx_builder) => Some(wasi_ctx_builder.build_context(ruby)?),
+
+      
+        let kw = scan_args::get_kwargs::<_, (), (Option<&WasiCtxBuilder>, Option<&WasiDeterministicCtxBuilder>), ()>(
+            args.keywords,
+            &[],
+            &[*WASI_CTX, *WASI_DET_CTX],
```
cameronbarker, line 128: There is a better way to do this, instead of creating addition kwargs. But I'd like to confirm it's working in flow-runtime and then will focus on refactoring.

ext/src/ruby_api/store.rs
```diff
@@ -184,6 +192,10 @@ impl Store {
         unsafe { &mut *self.inner.get() }.set_epoch_deadline(ticks_beyond_current);
     }
 
+    pub fn has_wasi_ctx(&self) -> bool {
```
cameronbarker, line 195: Added this to allow fuller test coverage.

ext/src/ruby_api/wasi_deterministic_ctx_builder.rs
```diff
@@ -0,0 +1,202 @@
+use super::root;
```
cameronbarker, line 1: This and `wasi_ctx_builder` have many similarities. When refactoring see if there is a way to merge these.

PR #277, opened 2024-01-10
Unlock the GVL when compiling WASM code

ianks:
Compiling code is CPU intensive, and can often take a significant amount of time (100ms+). Currently, we hold the GVL while compiling which means that no other Ruby threads can execute while compiling WASM. This makes gem a non-starter for multi-threaded apps (like Puma web apps).

### What this PR does

This PR makes it so we unlock the GVL in the following methods:

- `Wasmtime::Module.new`
- `Wasmtime::Module.from_file`
- `Wasmtime::Engine#precompile_file`

### DIFF
--- a/ext/src/helpers/mod.rs
+++ b/ext/src/helpers/mod.rs
@@ -1,6 +1,11 @@
 mod macros;
+mod nogvl;
 mod static_id;
 mod symbol_enum;
+mod tmplock;
 
+// pub use nogvl::without_gvl;
+pub use nogvl::nogvl;
 pub use static_id::StaticId;
 pub use symbol_enum::SymbolEnum;
+pub use tmplock::Tmplock;
--- a/ext/src/helpers/mod.rs
+++ b/ext/src/helpers/mod.rs
@@ -4,7 +4,6 @@ mod static_id;
 mod symbol_enum;
 mod tmplock;
 
-// pub use nogvl::without_gvl;
 pub use nogvl::nogvl;
 pub use static_id::StaticId;
 pub use symbol_enum::SymbolEnum;
--- a/ext/src/helpers/nogvl.rs
+++ b/ext/src/helpers/nogvl.rs
@@ -1,0 +1,29 @@
+use std::{ffi::c_void, mem::MaybeUninit, ptr::null_mut};
+
+use rb_sys::rb_thread_call_without_gvl;
+
+unsafe extern "C" fn call_without_gvl<F, R>(arg: *mut c_void) -> *mut c_void
+where
+    F: FnMut() -> R,
+    R: Sized,
+{
+    let arg = arg as *mut (&mut F, &mut MaybeUninit<R>);
+    let (func, result) = unsafe { &mut *arg };
+    result.write(func());
+
+    null_mut()
+}
+
+pub fn nogvl<F, R>(mut func: F) -> R
+where
+    F: FnMut() -> R,
+    R: Sized,
+{
+    let result = MaybeUninit::uninit();
+    let arg_ptr = &(&mut func, &result) as *const _ as *mut c_void;
+
+    unsafe {
+        rb_thread_call_without_gvl(Some(call_without_gvl::<F, R>), arg_ptr, None, null_mut());
+        result.assume_init()
+    }
+}
--- a/ext/src/helpers/tmplock.rs
+++ b/ext/src/helpers/tmplock.rs
@@ -1,0 +1,45 @@
+use magnus::{
+    rb_sys::{protect, AsRawValue},
+    RString,
+};
+
+pub trait Tmplock {
+    unsafe fn as_locked_slice(&self) -> Result<(&[u8], TmplockGuard), magnus::Error>;
+    unsafe fn as_locked_str(&self) -> Result<(&str, TmplockGuard), magnus::Error>;
+}
+
+#[derive(Debug)]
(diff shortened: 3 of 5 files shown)


ext/src/ruby_api/engine.rs
```diff
@@ -168,8 +171,9 @@ impl Engine {
     /// @return [String] Binary String of the compiled module.
     /// @see Module.deserialize
     pub fn precompile_module(&self, wat_or_wasm: RString) -> Result<RString, Error> {
-        self.inner
-            .precompile_module(unsafe { wat_or_wasm.as_slice() })
+        let (wat_or_wasm, _guard) = unsafe { wat_or_wasm.as_locked_slice() }?;
```
jbourassa, line 174: First time I come across [`rb_str_locktmp`](https://github.com/ruby/ruby/blob/31371b2e24b03ccb0a03b622faf8c65e6cf6a31a/include/ruby/internal/intern/string.h#L615) (thanks!).

IIUC:
- the lock ensures the string won't be mutated by another Ruby thread
- the RString's value is on the stack so it won't get GC'd nor moved/compacted

Do I get this right?

ext/src/helpers/mod.rs
```diff
@@ -1,6 +1,11 @@
 mod macros;
+mod nogvl;
 mod static_id;
 mod symbol_enum;
+mod tmplock;
 
+// pub use nogvl::without_gvl;
```
jbourassa, line 7:
```suggestion
```

jbourassa (commented):
I'll need a bit more time to review as I learn about those Ruby APIs but at a first look that makes sense to me.

ext/src/ruby_api/engine.rs
ianks: Yes and won't be "tampered with" (according to docs)

>  Obtains a "temporary  lock" of the string.  This  advisory locking mechanism
> prevents other  cooperating threads from  tampering the receiver.   The same
> thing could be done via freeze mechanism,  but this one can also be unlocked
> using rb_str_unlocktmp().

ext/src/helpers/nogvl.rs
```diff
@@ -0,0 +1,29 @@
...
+pub fn nogvl<F, R>(mut func: F) -> R
+where
+    F: FnMut() -> R,
+    R: Sized,
+{
+    let result = MaybeUninit::uninit();
+    let arg_ptr = &(&mut func, &result) as *const _ as *mut c_void;
+
+    unsafe {
+        rb_thread_call_without_gvl(Some(call_without_gvl::<F, R>), arg_ptr, None, null_mut());
```
jbourassa, line 26: re `None` ubf, [according to the doc](https://github.com/ruby/ruby/blob/3fe2f2689f0074a4c880218ca08cb10137880760/thread.c#L1577-L1579):

> Be careful to use rb_thread_call_without_gvl(). If you don't provide proper ubf(), your program will not stop for Control+C or other shutdown events.

For compilation, that seems fine: (1) I don't know of a way to interrupt compilation (unsure) (2) that's no different than today's behaviour.

jbourassa, line 26: I'm also unsure on signal handling, the doc has warnings about that. I'm not knowledgeable enough in that area to know if it's a concern.

I tried compiling in a tight loop and hitting Control-C and that seemed to work "fine", or at least, no differently than main.

If you want to share your thoughts on that, please do, but I think it's fine?

jbourassa (approved):
I left a few more questions / thoughts but no major concerns.

ext/src/ruby_api/module.rs
```diff
@@ -83,8 +95,10 @@ impl Module {
     /// @return [String]
     /// @see .deserialize
     pub fn serialize(&self) -> Result<RString, Error> {
```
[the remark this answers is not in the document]
ianks: Yes you're probably right here. Probably not worth the nogvl for serialization, just compilation.

ext/src/helpers/nogvl.rs
ianks: There's no way to interrupt compilation safely, so _during_ compilation interrupts will be ignored which is OK I think

ext/src/helpers/tmplock.rs
```diff
@@ -0,0 +1,45 @@
+use magnus::{
+    rb_sys::{protect, AsRawValue},
+    RString,
+};
+
+pub trait Tmplock {
+    unsafe fn as_locked_slice(&self) -> Result<(&[u8], TmplockGuard), magnus::Error>;
+    unsafe fn as_locked_str(&self) -> Result<(&str, TmplockGuard), magnus::Error>;
```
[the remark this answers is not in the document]
ianks: TBH, I wasn't 1000% sure that tmplocking the string would prevent any modifications of the string in libruby, so I opted for `unsafe`. Happy to reconsider that though 🤔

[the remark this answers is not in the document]
jbourassa: I guess we're not guaranteed it won't happen if a misbehaving C extension changes the buffer directly. I don't know if that's a valid reason for marking it unsafe, but I'd think not?

From libruby itself, looks like nothing should change the string, outside of a bug / oversight.

Clearly, I'm not 1000% sure either, but I'm inclined to mark this as safe. Your call though, I'm fine either way.

PR #279, opened 2024-01-10
Introducing Deterministic  Configuration to WasiCtxBuilder

cameronbarker:
A new approach based on #268.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,4 +100,4 @@ DEPENDENCIES
   yard-rustdoc (~> 0.3.2)
 
 BUNDLED WITH
-   2.5.4
+  2.5.4
--- a/ext/src/ruby_api/wasi_ctx_builder.rs
+++ b/ext/src/ruby_api/wasi_ctx_builder.rs
@@ -6,7 +6,9 @@ use magnus::{
 };
 use std::cell::RefCell;
 use std::{fs::File, path::PathBuf};
+use wasmtime_wasi::{preview2::bindings::wasi, stdio};
 use wasi_common::pipe::ReadPipe;
+
 
 enum ReadStream {
     Inherit,
@@ -44,6 +46,7 @@ struct WasiCtxBuilderInner {
     stderr: Option<WriteStream>,
     env: Option<Opaque<RHash>>,
     args: Option<Opaque<RArray>>,
+    deterministic: Option<bool>,
 }
 
 impl WasiCtxBuilderInner {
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -17,4 +17,5 @@ end
 
 group :test do
   gem "rspec", "~> 3.1"
+  gem "pry"
```
cameronbarker, line 20: Need to remove.

Gemfile.lock
```diff
@@ -100,4 +106,4 @@ DEPENDENCIES
   yard-rustdoc (~> 0.3.2)
 
 BUNDLED WITH
-   2.5.4
+   2.4.18
```
cameronbarker, line 109: Need to fix

ext/src/ruby_api/wasi_ctx_builder.rs
```diff
@@ -192,38 +195,56 @@ impl WasiCtxBuilder {
         rb_self
     }
 
+    pub fn set_deterministic(rb_self: RbSelf) -> RbSelf {
```
cameronbarker, line 198: I felt that this name might be incorrect given that it doesn't take a parameter. I had though potentially `as_deterministic`, but would appreciate any recommendations.

ext/src/ruby_api/wasi_ctx_builder.rs
```diff
@@ -192,38 +195,56 @@ impl WasiCtxBuilder {
...
+        inner.deterministic = Some(true);
+        rb_self
+    }
+
     pub fn build_context(&self, ruby: &Ruby) -> Result<wasmtime_wasi::WasiCtx, Error> {
-        let mut builder = wasmtime_wasi::WasiCtxBuilder::new();
         let inner = self.inner.borrow();
 
+        let mut wasi_ctx = if inner.deterministic.is_none() {
+            let mut ctx = wasmtime_wasi::WasiCtxBuilder::new();
```
cameronbarker, line 208: I had to make the ctx mutable to allow overriding of `push_args` and `push_env`

spec/fixtures/deterministic.rs
```diff
@@ -0,0 +1,85 @@
+//
```
cameronbarker, line 1: I included this file for:
1. Confirmation I'm writing code that would catch deterministic clock and random principals.
2. Ease of change for future updates.

saulecabrera (commented):
Thinking a bit more about this, given that the `WasiCtxBuilder` and the deterministic version of it operate on different levels:
* The builder api operates on a `WasiCtxBuilder` and creates a `WasiCtx`
* The only way of creating a  `WasiCtx`, is through `WasiCtx::new` (which is what the deterministic crate uses).

I was thinking that it might not be totally crazy to introduce a proper `WasiCtx` class that's exposed to Ruby. This approach will reflect more accurately the reality of the Rust entities at play and will also make it easier to:

* Create a deterministic version
* Give Ruby users a way to interact with the `WasiCtx` class more freely (currently there's no way of doing this, the only entity that you can interact with is the `WasiCtxBuilder`)

Once we have this class exposed, we could have an API, that is something like:

```ruby
class WasiCtx
  def deterministic(...) # The deterministic version
  def set_* # All (or a subset of) the setters exposed by the Rust API
end
```

One of the disadvantages that I see with the current approach is that it mixes the `WasiCtx` and `WasiCtxBuilder` concepts, which might get tricky to manage if we ever end up adding more methods to the builder.

PR #281, opened 2024-01-11
Add support for resource limits on `Wasmtime::Store`

ryanische:
This PR adds support to set limits in the constructor of a `Wasmtime::Store` using a `StoreLimitsBuilder`. 

Example usage:
```ruby
store = Store.new(engine, limits: {memories: 1, tables: 1})
```

### DIFF
--- a/ext/src/ruby_api/store.rs
+++ b/ext/src/ruby_api/store.rs
@@ -1,7 +1,7 @@
 use super::errors::wasi_exit_error;
 use super::{caller::Caller, engine::Engine, root, trap::Trap, wasi_ctx_builder::WasiCtxBuilder};
 use crate::{define_rb_intern, error};
-use magnus::Class;
+use magnus::value::StaticSymbol;
 use magnus::{
     class, function,
     gc::{Compactor, Marker},
@@ -10,6 +10,7 @@ use magnus::{
     value::Opaque,
     DataTypeFunctions, Error, IntoValue, Module, Object, Ruby, TypedData, Value,
 };
+use magnus::{Class, RHash};
 use std::cell::UnsafeCell;
 use std::convert::TryFrom;
 use wasmtime::{
@@ -20,11 +21,7 @@ use wasmtime_wasi::{I32Exit, WasiCtx};
 
 define_rb_intern!(
     WASI_CTX => "wasi_ctx",
-    MEMORY_SIZE => "memory_size",
-    TABLE_ELEMENTS => "table_elements",
-    INSTANCES => "instances",
-    TABLES => "tables",
-    MEMORIES => "memories",
+    LIMITS => "limits",
 );
 
 pub struct StoreData {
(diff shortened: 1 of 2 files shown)


ext/src/ruby_api/store.rs
```diff
@@ -170,6 +172,64 @@ impl Store {
...
+        >(
+            args.keywords,
+            &[],
+            &["memory_size", "table_elements", "instances", "tables", "memories"]
+        )?;
+
+        let mut limiter: StoreLimitsBuilder = StoreLimitsBuilder::new();
+        let (memory_size, table_elements, instances, tables, memories) = kw.optional;
+
+        if memory_size.is_some() {
```
ryanische, line 204: Not sure if there's a more elegant/idiomatic way to deal with all of these `Option` variables.

ext/src/ruby_api/store.rs
```diff
@@ -170,6 +172,64 @@ impl Store {
...
+        let kw = scan_args::get_kwargs::<
+            _,
+            (),
+            (Option<u64>, Option<u64>, Option<u64>, Option<u64>, Option<u64>),
+            ()
+        >(
+            args.keywords,
+            &[],
+            &["memory_size", "table_elements", "instances", "tables", "memories"]
+        )?;
```
ryanische, line 199: This line became quite long so I broke it up - not sure if there's a preferred style here.

ext/src/ruby_api/store.rs
saulecabrera: To avoid the the `is_some()` + `unwrap` combination, I'd suggest using the [`if let` pattern](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html), that's probably a bit more idiomatic.

ext/src/ruby_api/store.rs
saulecabrera: We usually let cargo format decide the formatting, so you could use 
```
cargo clippy -- -D warnings && cargo fmt --check
```

To check any clippy or format errors, and `cargo fmt` will format the codebase for you.

ext/src/ruby_api/store.rs
```diff
@@ -170,6 +172,64 @@ impl Store {
...
+            limiter = limiter.table_elements(table_elements.unwrap() as u32);
+        }
+
+        if instances.is_some() {
+            limiter = limiter.instances(instances.unwrap() as usize);
+        }
+
+        if tables.is_some() {
+            limiter = limiter.tables(tables.unwrap() as usize);
+            println!("SET TABLES");
```
saulecabrera, line 218:
```suggestion
            
```

ext/src/ruby_api/store.rs
```diff
@@ -170,6 +172,64 @@ impl Store {
...
+    /// @param table_elements [Integer]
+    ///   The maximum number of elements in a table.
+    /// @param instances [Integer]
+    ///   The maximum number of instances that can be created for a Store.
+    /// @param tables [Integer]
+    ///   The maximum number of tables that can be created for a Store.
+    /// @param memories [Integer]
+    ///   The maximum number of linear memories that can be created for a Store.
+    /// @raise [Error] if arguments can not be converted.
+    pub fn set_limits(&self, args: &[Value]) -> Result<(), Error> {
```
saulecabrera, line 188: Even though we had discussed using `wasmtime-py`'s approach here, upon a deeper look I do wonder if it'll be better if instead of using the `set_limiter` method, we allow the limits to be optionally passed when creating the store, (in `Store#new`) similar to how the Wasi context is passed in. This will make this more uniform with the current API IMHO. I _think_ that `wasmtime-py` convention is to favor the `set_*` approach for most of the APIs (e.g. it exposes a `set_wasi` method for example).

ryanische: I've updated this to set the limits in the constructor - I'm not sure whether there is a simpler way to call the `limiter` method from within the constructor and whether `StoreData.store_limits` is necessary any more.

saulecabrera (approved):
LGTM! Thanks for iterating on this.

PR #282, opened 2024-01-13
Expose Wasi Context to Ruby

cameronbarker:
A new approach based on the feedback from @saulecabrera -> https://github.com/bytecodealliance/wasmtime-rb/pull/279#pullrequestreview-1816008613

Our project would like to introduce `WasiCtx` that ensures that Clock and Random return deterministic results. To do so, we've exposed `WasiCtx` with a deterministic initializer to Ruby.  The deterministic instance is generated using  [Shopify/deterministic-wasi-ctx](https://github.com/Shopify/deterministic-wasi-ctx) crate. 

### Usage
```ruby
engine = Wasmtime::Engine.new
mod = Wasmtime::Module.from_file(engine, "spec/fixtures/wasi-debug.wasm")
linker = Wasmtime::Linker.new(engine, wasi: true)

# Wasi Deterministic Ctx
wasi_ctx = Wasmtime::WasiCtx.deterministic
store = Wasmtime::Store.new(engine, wasi_ctx: wasi_ctx)

instance = linker.instantiate(store, mod)
instance.invoke("_start")
```

## Breaking Changes
This PR changes `Store` `wasi_ctx` parameter to now accept `WasiCtx` instances. This will require WasiCtxBuild to return a `WasiCtx` instance now. To do so, we've exposed a `build` method that will compile the `WasiCtxBuild` instance and return a compiled `WasiCtx`.

**Pre v16**
```ruby
wasi_ctx = Wasmtime::WasiCtxBuilder.new # => WasiCtxBuilder
store = Wasmtime::Store.new(engine, wasi_ctx: wasi_ctx)
```

**v17+**
```ruby
wasi_ctx = Wasmtime::WasiCtxBuilder.new.build # => WasiCtx
store = Wasmtime::Store.new(engine, wasi_ctx: wasi_ctx)
```


ext/src/ruby_api/wasi_context.rs
```diff
@@ -0,0 +1,56 @@
+use super::{root, WasiCtxBuilder, wasi_ctx_builder::{file_r, file_w, wasi_file}};
+use crate::error;
+use std::{fs::File, path::PathBuf, cell::RefCell};
+use magnus::{class, method, function, prelude::*, RTypedData, Error, Ruby, Value, RString};
+use wasmtime_wasi::WasiCtx as WasiCtxImpl;
+use deterministic_wasi_ctx::build_wasi_ctx as wasi_deterministic_ctx;
+use wasi_common::pipe::ReadPipe;
+
+#[magnus::wrap(class = "Wasmtime::WasiContext")]
```
saulecabrera, line 9: I think we should also consider adding `#[magnus::wrap(class = "Wasmtime::WasiContext", size, free_immediately)]`; `size` and `free_immediately`

Ref: https://docs.rs/magnus/latest/magnus/attr.wrap.html

spec/fixtures/deterministic.rs
```diff
@@ -0,0 +1,85 @@
...
+//      - UTC time after sleep: utc2
+//      - System time before sleep: system_time1
+//      - System time after sleep: system_time2
+//      - Elapsed time: system_time1_elapsed
+//
+//  stderr: "Error: This is an error message"
+//
+// Contributing Notes:
+//  Compile: `cargo build --target wasm32-wasi`
+//  Run: `wasmer run target/wasm32-wasi/debug/deterministic.wasm`
```
saulecabrera, line 22:
```suggestion
//  Run: `wasmtime run target/wasm32-wasi/debug/deterministic.wasm`
```

ext/src/ruby_api/wasi_context.rs
saulecabrera: I also wonder if this should be exposed as `WasiCtx` to Ruby to reflect the actual naming in the Rust API. 

https://docs.wasmtime.dev/api/wasmtime_wasi/struct.WasiCtx.html

saulecabrera (commented):
The general direction looks good to me. I added some questions and suggestions inline. More generally I think it'd make sense to start adding the all the yard docs to `WasiCtx`.

ext/src/ruby_api/wasi_ctx.rs
```diff
@@ -0,0 +1,110 @@
...
+#[magnus::wrap(class = "Wasmtime::WasiCtx", size, free_immediately)]
+pub struct WasiCtx {
+    inner: RefCell<WasiCtxImpl>,
+}
+
+type RbSelf = Obj<WasiCtx>;
+
+impl WasiCtx {
+    /// @yard
+    /// Create a new {WasiCtx} with determinisitic Clock and Random. By default, it has nothing: no stdin/out/err.
```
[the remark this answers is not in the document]
cameronbarker: 👍

ext/src/ruby_api/wasi_ctx_builder.rs
```diff
@@ -192,9 +192,10 @@ impl WasiCtxBuilder {
         rb_self
     }
 
-    pub fn build_context(&self, ruby: &Ruby) -> Result<wasmtime_wasi::WasiCtx, Error> {
+    pub fn build(rb_self: RbSelf) -> Result<WasiCtx, Error> {
```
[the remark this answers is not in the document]
cameronbarker: Changed. Strangely, it couldn't be put into the second position without causing an error. But if I put ruby as the first parameter and self as the second, there were no errors.

saulecabrera (approved):
Thanks for iterating on this!

PR #285, opened 2024-01-29
chore: Update to wasmtime v17

saulecabrera:


ianks (approved):

ianks:
Seems like ruby-head may be busted, it's a moving target so fine to ignore

saulecabrera:
@ianks Thanks for the heads up. I was thinking, would you be fine if we softened the CI criteria for Ruby head? Emit a warning instead of an error if there are failures?

saulecabrera:
I'll merge this as is and open a new issue to discuss options for Ruby head.

saulecabrera:
https://github.com/bytecodealliance/wasmtime-rb/issues/286, feel free to leave any thoughts there and I'm happy to follow up with a PR once/if we settle on a particular solution.

Issue #286, opened 2024-01-30
Soften checks for Ruby head in CI


saulecabrera:
There have been several instances ([the most recent one](https://github.com/bytecodealliance/wasmtime-rb/pull/285)) where Ruby's head version fails in CI. This is not always ideal, given that most of the time the failures are not directly related to changes in a particular pull-request, instead the failures are related to changes in Ruby's latest version. If there's interest in keeping checks against Ruby's head version, would it be reasonable to "soften" these checks and emit a warning instead of an error?

PR #287, opened 2024-01-30
chore: Disable Ruby head in CI temporarily

saulecabrera:
The failure in main is preventing a release of v17 of the gem, so until we reach consensus about what to do with `head`, I think it's reasonable to exclude head for CI.

ref: https://github.com/bytecodealliance/wasmtime-rb/issues/286

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,12 @@ jobs:
     steps:
       - id: fetch
         uses: oxidize-rb/actions/fetch-ci-data@v1
+        with:
+          # TODO(#286)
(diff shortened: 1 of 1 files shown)


saulecabrera:
The release workflow is already ignoring `head`, so probably something else is going on.

saulecabrera:
I can confirm that gems are building successfully, but `actions/download-artifact` apparently can't find the generated files. 

```
Error: Unable to download artifact(s): Artifact not found for name: cross-gem
```

saulecabrera:
I'll close this for now, we can re-open it as part of https://github.com/bytecodealliance/wasmtime-rb/issues/286.

PR #295, opened 2024-02-08
upgrade wasmtime to 17.0.1

glebpom:

PR #296, opened 2024-02-08
chore: Disable Ruby head temporarily

saulecabrera:
Ruby head is still failing on pull requests so I'm opting to disable it until we decide what to do with it in
https://github.com/bytecodealliance/wasmtime-rb/issues/286.

Another instance https://github.com/bytecodealliance/wasmtime-rb/pull/295


ianks (approved):
Sounds good to me. I think `mswin` may be running head too but I'm not sure.

saulecabrera:
Seems to be, and we don't include `mswin` in the releases as far as I can tell, so I pushed an update that disables it too.

PR #317, opened 2024-04-25
chore: Update rb-sys

saulecabrera:
This commit updates rb-sys to the latest version. I'm hoping that this will fix the failures that are currently in main

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -20,7 +20,12 @@ jobs:
         uses: oxidize-rb/actions/fetch-ci-data@v1
         with:
           supported-ruby-platforms: |
-            exclude: [arm-linux] # no cranelift support yet
+            # Excluding:
+            # `arm-linux`: Cranelift doesn't support 32-bit architectures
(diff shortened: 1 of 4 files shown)


saulecabrera:
@ianks I'm mostly flying blind here, but do you think that this change will help resolve the failures in `main`: https://github.com/bytecodealliance/wasmtime-rb/actions/runs/8816496220/job/24246997925?

saulecabrera:
Ah, I'll confirm by running the action with this branch.

ianks:
yes 0.9.97 should fix that issue, as well add support for `aarch64-linux-musl`

saulecabrera:
Thanks, yeah that fixed it, there's also an issue with `x64-mingw32` I believe we need to exclude it give that 3.0 is no longer considered stable? See https://github.com/bytecodealliance/wasmtime-rb/pull/317/commits/68fb3d5b1de607d85123e6f2f35f67c9b36fb42c

saulecabrera:
Yeah, excluding `x64-mingw32` fixed the issue https://github.com/bytecodealliance/wasmtime-rb/actions/runs/8834566920

jbourassa (approved):

PR #322, opened 2024-05-05
Update to wasmtime v19

saulecabrera:
This commit reworks the bindings to support `wasmtime` v19 which introduces significant changes regarding reference types, namely `ValType` adds support for typed function refereces and GC. It's important to note that given that these features are not enabled by default, these proposals are not visible in `wasmtime-rb`'s API surface. It's reasonable to add support for them once the implementation in Wasmtime proper is enabled by default, however, this commit paves the road in a couple of places to support the `Ref` type.

Full changelog: https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#1900

### DIFF
--- a/ext/src/ruby_api/convert.rs
+++ b/ext/src/ruby_api/convert.rs
@@ -1,7 +1,7 @@
 use crate::{define_rb_intern, err, error, helpers::SymbolEnum};
 use lazy_static::lazy_static;
 use magnus::{prelude::*, Error, IntoValue, RArray, Ruby, Symbol, TryConvert, TypedData, Value};
-use wasmtime::{ExternRef, Val, ValType};
+use wasmtime::{ExternRef, RefType, Val, ValType};
 
 use super::{func::Func, global::Global, memory::Memory, store::StoreContextValue, table::Table};
 
@@ -13,6 +13,7 @@ define_rb_intern!(
     V128 => "v128",
     FUNCREF => "funcref",
     EXTERNREF => "externref",
+    NULLFUNCREF => "nullfuncref",
 );
 
 lazy_static! {
@@ -23,8 +24,8 @@ lazy_static! {
             (*F32, ValType::F32),
             (*F64, ValType::F64),
             (*V128, ValType::V128),
-            (*FUNCREF, ValType::FuncRef),
-            (*EXTERNREF, ValType::ExternRef),
+            (*FUNCREF, ValType::FUNCREF),
+            (*EXTERNREF, ValType::EXTERNREF),
         ];
 
         SymbolEnum::new("WebAssembly type", mapping)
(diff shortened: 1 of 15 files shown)


ext/src/ruby_api/convert.rs
```diff
@@ -122,22 +127,42 @@ impl ToExtern for Value {
 }
 
 pub trait ToSym {
-    fn to_sym(self) -> Symbol;
+    fn as_sym(&self) -> Result<Symbol, Error>;
```
saulecabrera, line 130: `ValType` is no longer `Copy`, which is the main motivation for this change. I also suspect that once APIs for typed function reference and GC are introduced it's likely that we will have to rework the relationship between `ValType` and its Ruby representation.

ext/src/ruby_api/params.rs
```diff
@@ -10,9 +10,6 @@ struct Param {
     index: u32,
     ty: ValType,
 }
-// Keep `Param` small so copying it to the stack is cheap, typically anything
-// less than 3usize is good
-assert_eq_size!(Param, [u64; 2]);
```
saulecabrera, line 15: `ValType` is not `Copy` anymore, so I don't think this assert is useful anymore, but let me know if there's anything that I'm missing here cc @ianks

saulecabrera:
`macos-latest, 3.1` is reporting a dynamic linking error 

    dyld[10102]: missing symbol called

I'm not entirely sure what's going on, however, the symptoms look very similar to https://github.com/oxidize-rb/rb-sys/pull/157

ext/src/ruby_api/convert.rs
jbourassa: What's your rationale for the naming change?

`#to_sym` is the normal name for converting to a Symbol, so I think the name could stay the same even if it can now raise.

ext/src/ruby_api/convert.rs
```diff
@@ -64,27 +64,32 @@ pub trait ToWasmVal {
...
-                let func_ref_value = match self.is_nil() {
-                    true => None,
-                    false => Some(*<&Func>::try_convert(*self)?.inner()),
-                };
-                Ok(Val::FuncRef(func_ref_value))
-            }
             ValType::V128 => err!("converting from Ruby to v128 not supported"),
+            // TODO: to be filled in once typed function references and/or GC
+            // are enabled by default.
+            t => err!("unsupported type: {t:?}"),
```
jbourassa, line 92: Nit / suggestion: I think it reads nicely when the RefType is a nested match:

```ruby
impl ToWasmVal for Value {
    fn to_wasm_val(&self, ty: ValType) -> Result<Val, Error> {
        match ty {
            ValType::I32 => Ok(i32::try_convert(*self)?.into()),
            ValType::I64 => Ok(i64::try_convert(*self)?.into()),
            ValType::F32 => Ok(f32::try_convert(*self)?.into()),
            ValType::F64 => Ok(f64::try_convert(*self)?.into()),
            ValType::V128 => err!("converting from Ruby to v128 not supported"),
            ValType::Ref(ty) => {
                if ty.matches(&RefType::EXTERNREF) {
                    let extern_ref_value = match self.is_nil() {
                        true => None,
                        false => Some(ExternRef::new(ExternRefValue::from(*self))),
                    };

                    Ok(Val::ExternRef(extern_ref_value))
                } else if ty.matches(&RefType::FUNCREF) {
                    let func_ref_value = match self.is_nil() {
                        true => None,
                        false => Some(*<&Func>::try_convert(*self)?.inner()),
                    };

                    Ok(Val::FuncRef(func_ref_value))
                } else {
                    // TODO: to be filled in once typed function references and/or GC
                    // are enabled by default.
                    err!("unsupported type: {ty:?}")
                }
            }
        }
    }
}
```

Same below. Use this version if you like it too, or disregard if you don't!

ext/src/ruby_api/func.rs
```diff
@@ -127,25 +132,25 @@ impl<'a> Func<'a> {
...
-            .results()
-            .map(ToSym::to_sym)
-            .collect();
-        Ok(results)
+        let ty = self.inner.ty(self.store.context()?);
+        let len = ty.results().len();
+        let mut results = ty.results();
+        results.try_fold(RArray::with_capacity(len), |array, r| {
+            array.push(r.as_sym()?).map_err(|e| error!("{e}"))?;
+            Ok::<_, Error>(array)
```
jbourassa, line 152:
```suggestion
            array.push(r.as_sym()?)?;
            Ok(array)
```


1. if `Array#push` fails (it should never because it's not frozen), we wouldn't want to raise a `Wasmtime::Error` but let the original error bubble, so I don't think we want the `map_err`.
2. Looks like the compiler figured out the return type just fine with `Ok(array)` on my side.

Same for params.

ext/src/ruby_api/params.rs
jbourassa: It may still have a small performance impact (we can benchmark it) but if that's the case we can come back to optimize it, it doesn't sound like a big deal?

jbourassa (approved):

ext/src/ruby_api/convert.rs
saulecabrera: The name change was mostly to reflect `owned` -> `borrow` in the signature, but then I realized that I misinterpreted this table https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Reverted!

ext/src/ruby_api/params.rs
saulecabrera: I don't think it's a big deal to be honest, but we can always come back later if this ends up being an issue.

ext/src/ruby_api/func.rs
saulecabrera: Fixed!

ext/src/ruby_api/params.rs
ianks: What's the new size of it? Can we assert on the new size instead to avoid accidentally bloating it?

ianks:
You can disable the rust-crate example on ruby 3.1, no idea what's going on there and why it's now failing 🤔

ianks (approved):

PR #354, opened 2024-08-13
Remove wasmtime-runtime by using resource limiter to track memory

jeffcharles:
I reached out to the Wasmtime maintainers a couple weeks ago about the wasmtime-runtime removal and they suggested looking into using a `ResourceLimiter` to track the memory usage. This is what I was able to come up with. I ensure any store created through the Ruby API has a limiter attached that forwards results from the underlying limiter but also informs the Ruby GC of allocations and deallocations. Any thoughts on the approach? In particular, what am I missing?

I've confirmed the `current` argument always starts with a value of `0` so it's not missing the initial allocation.


ext/src/ruby_api/store.rs
```diff
@@ -363,3 +366,70 @@ pub fn init() -> Result<(), Error> {
...
+    fn tables(&self) -> usize {
+        self.inner.tables()
+    }
+
+    fn memories(&self) -> usize {
+        self.inner.memories()
+    }
+}
+
+impl Drop for TrackingResourceLimiter {
```
saulecabrera, line 431: I believe you can rely on `ManuallyTracked`, which already takes care of doing this for you: https://docs.rs/rb-sys/latest/src/rb_sys/tracking_allocator.rs.html#195, similar to what was done in the original PR https://github.com/bytecodealliance/wasmtime-rb/pull/187/files#diff-b8e1c7a63c88d8176a489815defd575b01f6a596a617bb8561bb4fe54ac808d1R15

saulecabrera (commented):
Generally looks good to me; left a suggestion, which I think it makes sense to hash out before merging.

ext/src/ruby_api/store.rs
jeffcharles: Ah, I see `ManuallyTracked` has a `new` method (as opposed to `wrap` which requires an inner item). I can make that change.

saulecabrera (approved):

PR #364, opened 2024-08-19
Update to Wasmtime 23

jeffcharles:
Update to Wasmtime 23.0.2.

This enables the `memory-protection-keys` feature in Wasmtime so the pooling allocator configuration options that are gated by this feature can be used. Otherwise we would likely have to remove the API to enable using these configuration options. There haven't been any complaints about performance so keeping that logic in place _should_ be fine.

The tests for the pooling allocator config have also been updated since Wasmtime no longer rounds up the value to a page size at configuration time but does it at runtime instead (see [this commit message in Wasmtime for details](https://github.com/bytecodealliance/wasmtime/commit/bdd7842263c1840c328d4ffc2a53323c83af5203)).


jeffcharles:
I think the `build_doc` might be failing because we're using `nightly-2023-12-30` for that job and Wasmtime 23 requires at least Rust 1.77.0.

saulecabrera (approved):

PR #371, opened 2024-09-06
Add support to enable/disable wasm reference types

davejcameron:
Add support to enable/disable wasm reference types on engine config.


saulecabrera (commented):
Looks good to me, can you add the `reference_types` key in this test? https://github.com/bytecodealliance/wasmtime-rb/blob/main/spec/unit/engine_spec.rb#L31

saulecabrera (approved):

PR #372, opened 2024-09-06
Support fetching imports from a Module

davejcameron:
Add support for fetching the list of Module imports.

### DIFF
--- a/ext/src/ruby_api/externals.rs
+++ b/ext/src/ruby_api/externals.rs
@@ -44,7 +44,7 @@ impl ExternType {
     /// @yard
     /// Returns the exported function's FuncType or raises a `{ConversionError}` when the export is not a
     /// function.
-    /// @return [Func] The exported function.
+    /// @return [FuncType] The exported function's type.
     pub fn to_func_type(ruby: &Ruby, rb_self: Obj<Self>) -> Result<Value, Error> {
         match *rb_self {
             ExternType::Func(f) => Ok(f.as_value()),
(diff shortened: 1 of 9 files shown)


ext/src/ruby_api/module.rs
```diff
@@ -106,6 +109,24 @@ impl Module {
...
+    /// @yard
+    /// Returns the list of imports that this Module has and must be satisfied.
+    /// @return [Array<Hash>] An array of hashes containing import information
+    pub fn imports(&self) -> Result<RArray, Error> {
+        let module = self.get();
+        let imports = module.imports();
+
+        let result = RArray::new();
+        for import in imports {
+            let hash = RHash::new();
```
saulecabrera, line 122: The exports' length is known, I believe here we can use `RArray::with_capacity(imports.len())` to minimize the cost of reallocations. We can do the same with the usage of `RHash`.

ext/src/ruby_api/module.rs
```diff
@@ -106,6 +109,24 @@ impl Module {
...
+            hash.aset("module", import.module())?;
+            hash.aset("name", import.name())?;
+            hash.aset("type", format!("{:?}", import.ty()))?;
```
saulecabrera, line 125: I'm not sure we want to return a string for the type, ideally we'd want a proper type representing the extern. I believe you can use https://github.com/bytecodealliance/wasmtime-rb/blob/main/ext/src/ruby_api/externals.rs#L110

ext/src/ruby_api/externals.rs
```diff
@@ -1,13 +1,98 @@
...
+        }
+    }
+}
+unsafe impl Send for ExternType {}
+
+impl ExternType {
+    /// @yard
+    /// Returns the exported function's FuncType or raises a `{ConversionError}` when the export is not a
+    /// function.
+    /// @return [Func] The exported function.
```
saulecabrera, line 47: I believe here it should be

```
@return [FuncType]
```

In the signatures below as well?

saulecabrera (commented):
Looks good to me; there are couple of docs that need updating and after we should be able to land this one.

ext/src/ruby_api/externals.rs
mkcny: Good catch! Fixed.

PR #374, opened 2024-09-11
Add memory usage and memory limit hit

davejcameron:
Add two new methods to the Store to be able to check if the memory limit has been hit, and the maximum linear memory used.

### DIFF
--- a/ext/src/ruby_api/store.rs
+++ b/ext/src/ruby_api/store.rs
@@ -429,13 +429,6 @@ impl ResourceLimiter for TrackingResourceLimiter {
         desired: usize,
         maximum: Option<usize>,
     ) -> anyhow::Result<bool> {
-        if let Some(max) = maximum {
-            if desired > max {
-                self.linear_memory_limit_hit = true;
-                return Ok(false);
-            }
-        }
-
         let res = self.inner.memory_growing(current, desired, maximum);
 
         // Update max_linear_memory_consumed
--- a/ext/src/ruby_api/store.rs
+++ b/ext/src/ruby_api/store.rs
@@ -32,8 +32,6 @@ pub struct StoreData {
     refs: Vec<Value>,
     last_error: Option<Error>,
     store_limits: TrackingResourceLimiter,
-    linear_memory_limit_hit: bool,
-    max_linear_memory_consumed: usize,
 }
 
 impl StoreData {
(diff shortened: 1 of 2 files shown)


ext/src/ruby_api/store.rs
```diff
@@ -391,9 +468,24 @@ impl ResourceLimiter for TrackingResourceLimiter {
...
+            if desired > memory_size_limit {
+                self.linear_memory_limit_hit = true;
+                return Ok(false);
+            }
+        }
+
+        if res.is_ok() {
             self.tracker.increase_memory_usage(desired - current);
+        } else {
+            self.linear_memory_limit_hit = true;
```
saulecabrera, line 486: `self.inner.memory_growing` is [already checking](https://docs.rs/wasmtime/latest/src/wasmtime/runtime/limits.rs.html#334-352) the following invariants:

* `desired > limit` (which is the same `memory_size` that we're passing through the limits hash)
* `desired > max (if defined)`


With that in mind, I wonder if we can simplify this method to:

```rust
 // Update max_linear_memory_consumed
 self.max_linear_memory_consumed = self.max_linear_memory_consumed.max(desired);
       
if res.is_ok() {
  self.tracker.increase_memory_usage(desired - current);
} else {
  self.linear_memory_limit_hit = true;
}
res
```

ext/src/ruby_api/store.rs
```diff
@@ -390,10 +429,30 @@ impl ResourceLimiter for TrackingResourceLimiter {
         desired: usize,
         maximum: Option<usize>,
     ) -> anyhow::Result<bool> {
+        if let Some(max) = maximum {
+            if desired > max {
+                self.linear_memory_limit_hit = true;
+                return Ok(false);
+            }
+        }
```
saulecabrera, line 437: This check is already performed by the default implementation, (the `self.inner.memory_growing` call below) perhaps we can omit this, in general I'd say that it might make sense to leave the body of this function as:

```rust
if res.is_ok() {
    self.tracker.increase_memory_usage(desired - current);
 } else {
     self.linear_memory_limit_hit = true;
  }

   if let Ok(allowed) = res {
      if !allowed {
         self.linear_memory_limit_hit = true;
       }
   }
```

saulecabrera (approved):

PR #375, opened 2024-09-19
Fix YARD documentation

### LINKED ISSUE #342: `data_size` not rendered through YARD
Hey,

I have a feature request. In my work, I'm interested in knowing the memory usage of a WASM module. In particular, I want to expose https://docs.rs/wasmtime/latest/wasmtime/struct.Memory.html#method.data_size

I would simply have the same API in Ruby. Happy to contribute a PR.

jbourassa:
YARD doc generation is broken in a few different ways, this change aims to fix all the known issues. Fixes #342.

Note: this won't retrigger doc generation for existing releases, but once this is merged I can manually build and push the docs for the current release.

**Issues:**

1. Docs for tagged releases (vx.y.z tags) was last generated for v6.0.0 because we stopped running CI in on versions in #120. This means that the current "latest" doc symlink points to v6.0.0, helping other doc issues go unnoticed (see next points).

   Fix: run the doc generation step on release as well.

2. YARD doc generation stopped working at some point last year when we updated the nightly version. The doc for main only has a handful of Ruby-defined APIs, but none of the Rust-defined APIs. The root cause is a change in rustdoc's JSON format which caused yard-rustdoc to miss all definitions.

   Fix: teach yard-rustdoc about the new format, and update the dependency. This will also fix an issue causing `Wasmtime::Memory::UnsafeSlice` to not show up in the class list sidebar. See: https://github.com/oxidize-rb/yard-rustdoc/compare/v0.3.2...v0.4.0

3. Documentation entries had issues (invalid names, missing args, etc.) This would normally be caught in CI, but they weren't given the docs for Rust-defined APIs were skipped (see above point).

-------

Before & after:
Screenshot 2024-09-19 at 11 51 13 AM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,3 +128,6 @@ jobs:
           omitNameDuringUpdate: true
           omitPrereleaseDuringUpdate: true
           skipIfReleaseExists: true
+
+      - name: Publish doc
+        uses: ./.github/actions/publish-doc
--- a/ext/src/ruby_api/wasi_ctx.rs
+++ b/ext/src/ruby_api/wasi_ctx.rs
@@ -40,8 +40,8 @@ impl WasiCtx {
 
     /// @yard
     /// Set stdin to read from the specified file.
-    /// @param path [String] The path of the file to read from.
     /// @def set_stdin_file(path)
+    /// @param path [String] The path of the file to read from.
     /// @return [WasiCtxBuilder] +self+
     fn set_stdin_file(rb_self: RbSelf, path: RString) -> RbSelf {
         let inner = rb_self.inner.borrow_mut();
@@ -52,8 +52,8 @@ impl WasiCtx {
 
     /// @yard
     /// Set stdin to the specified String.
-    /// @param content [String]
     /// @def set_stdin_string(content)
+    /// @param content [String]
     /// @return [WasiCtx] +self+
     fn set_stdin_string(rb_self: RbSelf, content: RString) -> RbSelf {
         let inner = rb_self.inner.borrow_mut();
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -128,3 +128,6 @@ jobs:
           omitNameDuringUpdate: true
           omitPrereleaseDuringUpdate: true
           skipIfReleaseExists: true
+
+      - name: Publish doc
+        uses: ./.github/actions/publish-doc
```
jbourassa, line 133: This is untested, but should work? It works for the CI workflow: https://github.com/bytecodealliance/wasmtime-rb/blob/b5139ef7565e3f4e9671387db5c79f661b112f9d/.github/workflows/ci.yml#L118-L120

ext/src/ruby_api/wasi_ctx.rs
```diff
@@ -40,8 +40,8 @@ impl WasiCtx {
 
     /// @yard
     /// Set stdin to read from the specified file.
-    /// @param path [String] The path of the file to read from.
     /// @def set_stdin_file(path)
+    /// @param path [String] The path of the file to read from.
```
jbourassa, line 44: Putting `@def` before `@param` isn't required, but reads more nicely.

saulecabrera (approved):
Thanks!

> The root cause is a change in rustdoc's JSON format which caused yard-rustdoc to miss all definitions.

Thanks for describing the root cause as well, I was never able to get to the bottom if it!

jbourassa:
> once this is merged I can manually build and push the docs for the current release.

Done: https://bytecodealliance.github.io/wasmtime-rb/latest/

PR #378, opened 2024-09-24
Build docs action

### LINKED ISSUE #377: `Publish doc` and `Post Publish doc` actions consistently failing
I haven't done a deep investigation here, however I've seen consistent failures in the `Publish doc` and `Post Publish doc` actions on main. I'm not entirely sure what's causing them. Here's a screenshot of the failure:

Screenshot 2024-09-20 at 5 16 02 PM

jbourassa:
In #375, I tried re-enabling doc generation on releases by calling the "publish doc" action in the release workflow. That didn't work: the "publish doc" action requires doc to be built beforehand in the CI workflow, but the CI workflow does not run on release.

I'm fixing it by moving the documentation generation from the CI workflow to its own workflow. The new doc workflow runs on push and commit, just like CI. It's desirable that doc generation runs on any commit to catch issues, e.g. this script that ensures documented constants and methods indeed exist: https://github.com/bytecodealliance/wasmtime-rb/blob/ae2e78da061ff5927b4a7a9acd0b0358c7d58926/rakelib/doc.rake#L12-L45

I tested the workflow by pushing new tags in a fork; it should work this time.

Fixes #377


rakelib/doc.rake
```diff
@@ -94,4 +94,4 @@ namespace :doc do
   end
 end
 
-task doc: ["compile", "doc:default"]
+task doc: ["env:dev", "compile", "doc:default"]
```
jbourassa, line 97: Unrelated to the issue but that shaved off a few minutes.

.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - uses: oxidize-rb/actions/setup-ruby-and-rust@v1
         with:
           ruby-version: "3.3"
-          bundler-cache: false
+          bundler-cache: true
```
jbourassa, line 46: Also unrelated, but that sped up things and I can't see a reason not to use it. It's set to true elsewhere.

saulecabrera (approved):

PR #391, opened 2024-10-25
Support components `result` types

jbourassa:
Add support for the component model's result type.

`result<O, E>` are mapped to instances of `Wasmtime::Component::Result`. I intentionally left the interface simple, but we can expand it if we see fit.

Some design choices:
- `Result` behave like value objects for equality and hashing, similar to hash, array, etc.
- `Result` is implemented in Ruby for simplicty.
- When converting Ruby `Result` to a result branch of a none type (e.g. unparametrized `result`), the Ruby value *must* be nil. This follows the same spirit that the Ruby bindings generally avoids type casting as a way to avoid subtle bugs.
  Example: converting `Result.ok(1)` to a component model `result<_, E>` will error. Note that `_` here means "unit".

-----

This PR also includes slight non-controversial tweaks to tests and conversion logic of records.


ext/src/ruby_api/component/convert.rs
```diff
@@ -164,9 +181,64 @@ pub(crate) fn rb_to_component_val(
...
+                        } else {
+                            err!(
+                                "expected nil for result<O, _> error branch, got {}",
+                                err_value.inspect()
+                            )
+                        }
+                    }
+                }
+            }
+        }
```
jbourassa, line 222: This is a mouthful. I tried simplifying but did not come up with anything much better. Ideas are welcome.

ext/src/ruby_api/component/convert.rs
```diff
@@ -164,9 +181,64 @@ pub(crate) fn rb_to_component_val(
...
+                        if ok_value.is_nil() {
+                            Ok(Val::Result(Ok(None)))
+                        } else {
+                            err!(
+                                "expected nil for result<_, E> ok branch, got {}",
+                                ok_value.inspect()
+                            )
+                        }
+                    }
+                }
```
saulecabrera, line 204: This invariant is the same in both cases (ok, error), perhaps this could be extracted into a local function?

saulecabrera (approved):
> When converting Ruby Result to a result branch of a none type (e.g. unparametrized result), the Ruby value must be nil. This follows the same spirit that the Ruby bindings generally avoids type casting as a way to avoid subtle bugs.
Example: converting Result.ok(1) to a component model result<_, E> will error. Note that _ here means "unit".

IMHO, this is an important design decision that should probably be documented somewhere in the public API/docs? Maybe the result class is a good place to start?

ext/src/ruby_api/component/convert.rs
jbourassa: Is this more or less what you had in mind?

```diff
@@ -185,38 +198,40 @@ pub(crate) fn rb_to_component_val(
             // Expect value to conform to `Wasmtime::Component::Value`'s interface
             let ruby = Ruby::get().unwrap();
             let is_ok = value.funcall::<_, (), bool>(IS_OK.into_id_with(&ruby), ())?;
+            let expect_nil = |value: Value,
+                              nil_case: Val,
+                              error_handler: fn(Value) -> Result<Val, Error>|
+             -> Result<Val, Error> {
+                if value.is_nil() {
+                    Ok(nil_case)
+                } else {
+                    error_handler(value)
+                }
+            };
 
             if is_ok {
                 let ok_value = value.funcall::<_, (), Value>(OK.into_id_with(&ruby), ())?;
                 match result_type.ok() {
                     Some(ty) => rb_to_component_val(ok_value, _store, &ty)
                         .map(|val| Val::Result(Result::Ok(Some(Box::new(val))))),
-                    None => {
-                        if ok_value.is_nil() {
-                            Ok(Val::Result(Ok(None)))
-                        } else {
-                            err!(
-                                "expected nil for result<_, E> ok branch, got {}",
-                                ok_value.inspect()
-                            )
-                        }
-                    }
+                    None => expect_nil(ok_value, Val::Result(Ok(None)), |v| {
+                        err!(
+                            "expected nil for result<_, E> ok branch, got {}",
+                            v.inspect()
+                        )
+                    }),
                 }
             } else {
                 let err_value = value.funcall::<_, (), Value>(ERROR.into_id_with(&ruby), ())?;
                 match result_type.err() {
                     Some(ty) => rb_to_component_val(err_value, _store, &ty)
                         .map(|val| Val::Result(Result::Err(Some(Box::new(val))))),
-                    None => {
-                        if err_value.is_nil() {
-                            Ok(Val::Result(Err(None)))
-                        } else {
-                            err!(
-                                "expected nil for result<O, _> error branch, got {}",
-                                err_value.inspect()
-                            )
-                        }
-                    }
+                    None => expect_nil(err_value, Val::Result(Err(None)), |v| {
+                        err!(
+                            "expected nil for result<O, _> error branch, got {}",
+                            v.inspect()
+                        )
+                    }),
                 }
             }
         }
```

I preferred the prior version; this one doesn't communicate the intent much better but adds indirection. If you can think of a better approach, please share 😄.

saulecabrera: I'm good keeping the previous version!

PR #399, opened 2024-11-14
Replace `Instance#invoke` with `Func#call`

jbourassa:
Following the rust crate API, `Instance#invoke` is removed and replaced with 2 calls: `Instance#get_func` and `Func#call`.

Other notable changes:
- Wasm component types <-> Ruby conversion is now documented under `Func`. I ran into Yard limitations when trying to style certain things, but I think it renders well enough:

type-conversion

- `Instance#get_func` supports nested exports by accepting an array of strings.

  ```ruby
  instance.get_func("foo") # looks up the `foo` export
  instance.get_func(["bar", "baz"]) # looks up the `bar` export, then `baz`
  ```

  This API can be extended later to accept an instance of `ExportIndex` to avoid the cost of comparing exports by name each time, like the Rust API. For example:
  
  ```ruby
  index = component.get_export("foo") # or get_export(["bar", "baz"])
  # => #<Wasmtime::Component::ComponentExportIndex ...>
  instance.get_func(index)
  ```

### DIFF
--- a/ext/src/ruby_api/component/func.rs
+++ b/ext/src/ruby_api/component/func.rs
@@ -1,35 +1,118 @@
-use super::convert::{component_val_to_rb, rb_to_component_val};
-use crate::ruby_api::{errors::ExceptionMessage, store::StoreContextValue};
-use magnus::{exception::arg_error, prelude::*, value, Error, IntoValue, RArray, Value};
+use crate::ruby_api::{
+    component::{
+        convert::{component_val_to_rb, rb_to_component_val},
+        Instance,
+    },
+    errors::ExceptionMessage,
+    store::{Store, StoreContextValue},
+};
+use magnus::{
+    class, exception::arg_error, gc::Marker, method, prelude::*, typed_data::Obj, value,
+    DataTypeFunctions, Error, IntoValue, RArray, RModule, Ruby, TypedData, Value,
+};
 use wasmtime::component::{Func as FuncImpl, Type, Val};
 
-pub struct Func;
+/// @yard
+/// @rename Wasmtime::Component::Func
+/// Represents a WebAssembly component Function
+/// @see https://docs.wasmtime.dev/api/wasmtime/component/struct.Func.html Wasmtime's Rust doc
+///
+/// == Component model types conversion
+///
+/// Here's how component model types map to Ruby objects:
+///
+/// bool::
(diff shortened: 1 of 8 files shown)


ext/src/ruby_api/component/func.rs
```diff
@@ -1,35 +1,118 @@
...
+///     Ruby +Integer+. Overflows raise.
+/// f32, f64::
+///     Ruby +Float+.
+/// string::
+///     Ruby +String+. Exception will be raised if the string is not valid UTF-8.
+/// list<T>::
+///     Ruby +Array+.
+/// tuple::
+///     Ruby +Array+ of the same size of tuple. Example: +tuple<T, U>+ would be converted to +[T, U]+.
+/// record::
```
saulecabrera, line 36: If I'm not wrong there was a valid reason not to use symbols, right? In case I'm remember this correctly, would it make sense to explicitly document it?

saulecabrera (approved):

ext/src/ruby_api/component/func.rs
jbourassa: The main reason is that creating a non-interned symbol is more work: you first need a `RString` and then [intern](https://github.com/ruby/ruby/blob/4bdfc9070cf626db4553a6cd15dbae19604180cf/symbol.c#L857-L894) it. Now I didn't benchmark (taking a note to do it).

Ideally we'd create symbols once and cache them for a record type, but I don't see how it can be done. We don't have anything that's "stable" to attach the symbols cache to. `wasmtime::component::types::Record` is specific to a given component.

The way to fix that, I think, would be to codegen for a WIT file, and convert to/from those codegen'd types.

Now that's long version of the explaination. The short version might be "for performance (see issue # for details)".

---

Thinking about it some more, I don't feel like string keys are a wart to the API. The JSON gem returns string keys by default, so does the postgre gem, etc. For anything that's looks dynamic, strings "feel" normal to me. Whether component model records are dynamic depends on how you look at it; a case could be made either way.

jbourassa: I created https://github.com/bytecodealliance/wasmtime-rb/issues/400 to track this. I'll go ahead and merge as is so that we get this API in before the next release, given Wasmtime 27 was just released.

PR #408, opened 2024-12-30
Ruby 3.4 update

jbourassa:
Support Ruby 3.4:

- Update rb-sys
- Fix tests (updated error message in Ruby 3.4)
- Fix example (old magnus was incompatible with 3.4)
- Update test and release workflows to use Ruby 3.4

-----

**TODOs:**
- [x] wait for https://github.com/oxidize-rb/rb-sys/pull/475
- [x] wait for https://github.com/ffi/ffi/pull/1133 and run `bundle` to re-add the ffi precompiled binaries to the lockfile
- [ ] Fix Windows compilation issue, see [this error](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/12551032834/job/34995930549?pr=408)
    
    **Investigation notes:**
    - Ruby 3.4 detects and uses `<stdckdint.h>` if it's present.
    - The version of llvm [rb-sys uses](https://github.com/oxidize-rb/rb-sys/blob/5d1999ba70d301581524b8252c8ff3b8a825afa2/docker/Dockerfile.x64-mingw32#L12) doesn't have stdckint (it was introduced in [v18.1.6](https://releases.llvm.org/18.1.6/tools/clang/docs/ReleaseNotes.html)).
    - looks like [the compiler thinks `<stdckdint.h>` is available](https://github.com/ruby/ruby/blob/3e8f0e558973aa8608e6d8da4dce33fe5dde28e1/include/ruby/internal/stdckdint.h#L30C11-L30C34), but then later fails to see it.

    ~Should be fixed once https://github.com/oxidize-rb/actions/pull/47 merges.~
    
    Edit: that's not the right fix unfortunately. I say we merge on red and accept Ruby 3.4+Windows failing CI for now. This will not affect pre-built binaries.

----

I won't cut a new version right away given we're due for a Wasmtime version bump, so we'll create a new release soon for that.


saulecabrera (approved):

jbourassa:
Thanks @saulecabrera. I have a PR cooking that should fix the Windows CI, I’ll open it later today.

saulecabrera:
There's [a failure on main](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/12727791787/job/35497579189), unrelated to Windows AFAICT;  I'm trying to sanity check if it's related to the Ruby upgrade or if it's due to something else. I haven't spent a bunch of time investigating this failure; have you seen this failure before?

jbourassa:
Hmm the issue is on 3.3, looks like something related to https://github.com/oxidize-rb/rb-sys/pull/475.

I downloaded the artifacts and 3.3 is indeed missing:

```
wasmtime-28.0.0-arm64-darwin ❯ tree lib
lib
├── wasmtime
│   ├── 3.1
│   │   └── wasmtime_rb.bundle
│   ├── 3.2
│   │   └── wasmtime_rb.bundle
│   ├── 3.4
│   │   └── wasmtime_rb.bundle
│   ├── component.rb
│   ├── error.rb
│   └── version.rb
└── wasmtime.rb
```

Now why aren't we the fix, that's the question.

jbourassa:
> Now why aren't we seeing the fix, that's the question.

I think it's related to how the GitHub action picks the version; it [uses the first `rb_sys` line in the `Gemfile.lock`](https://github.com/oxidize-rb/actions/blob/a075ffedfe329496d648c8690d8ef99ed3ac0fb1/cross-gem/action.yml#L44), regardless of the current version.

Updating the rb_sys minimum version in the gemspec should do it? Let's try that.

(Also, we're on a deprecated version of the cross-gem action, but hopefully that's not the culprit here).

jbourassa:
> (Also, we're on a deprecated version of the cross-gem action, ~but hopefully that's not the culprit here~ Edit: it may have something to do, more digging is necessary).

I've updated the workflow in my fork to use the current cross-gem action. That (with the rb-sys version bump?) should do it, but I'm running into a weird issue: the rustc process dies without any output:

```
/usr/local/cargo/bin/cargo rustc --crate-type cdylib --package wasmtime-rb --manifest-path /home/runner/work/wasmtime-rb/wasmtime-rb/ext/Cargo.toml --target aarch64-unknown-linux-gnu --target-dir /home/runner/work/wasmtime-rb/wasmtime-rb/tmp/rb-sys-dock/aarch64-linux/target --lib --profile release -- -C linker=aarch64-linux-gnu-gcc -L native=/usr/local/rake-compiler/ruby/aarch64-linux-gnu/ruby-3.1.0/lib -C link-arg=-pipe -C link-arg=-lm
   Compiling proc-macro2 v1.0.92
<snip...>
   Compiling postcard v1.1.1
Error: Process completed with exit code 1.
```

See [this example run](https://github.com/jbourassa/wasmtime-rb/actions/runs/12738502464/job/35501447372).

I can reproduce by SSHing into the GitHub Action and:
* editing the locally-installed gem source so that [`interactive?`](https://github.com/oxidize-rb/rb-sys/blob/241a1425681a96412002ab9cd992ea3e05e80585/gem/exe/rb-sys-dock#L341-L343) is false
* `rm -rf tmp` to ensure a fresh build
* `rb-sys-dock --platform aarch64-linux --directory . --ruby-versions 3.1,3.2,3.3,3.4 --build`

Interestingly, interactive or subsequent builds do not suffer from this.

I also found that not running [sigfw](https://github.com/rake-compiler/rake-compiler-dock/blob/c4e7dc390e0757891ad8c7898953f87a35c957dc/build/sigfw.c) as process 1 (again by editing the gem code) makes the build pass. Does that mean _something_ is sending SIGINT to our process while it compiles? To be continued.

jbourassa:
I created an issue with my finding and broader context: https://github.com/bytecodealliance/wasmtime-rb/issues/418

jbourassa:
Ruby 3.4 + Windows should now be green, CI is still running compilation and tests succeeded: https://github.com/bytecodealliance/wasmtime-rb/actions/runs/12750608774/job/35545705015

PR #416, opened 2025-01-08
Update to wasmtime v28

saulecabrera:


jbourassa:
I also tackled the build failures in https://github.com/bytecodealliance/wasmtime-rb/pull/408 -- my goal was to merge that one first.

It's stuck on windows CI failing on Ruby 3.4, but pre-built binaries still work. I investigated but not sure how to move forward, so we could allow Ruby 3.4 to fail on Windows until we get it fixed. Thoughts?

jbourassa (approved):

saulecabrera:
Sure let's merge https://github.com/bytecodealliance/wasmtime-rb/pull/408 first, happy to review when you feel it's ready. 

> It's stuck on windows CI failing on Ruby 3.4, but pre-built binaries still work. I investigated but not sure how to move forward, so we could allow Ruby 3.4 to fail on Windows until we get it fixed. Thoughts?

Just to make sure I fully understand: if we allow Ruby 3.4 to fail on CI, we won't need to exclude it from the pre-built binaries included in the release assets? Is that correct?

jbourassa:
Correct. The pre-built binaries use a docker image to build, they don’t run
on an actual Windows machine (unlike CI).

jbourassa (approved):

PR #417, opened 2025-01-13
Bump rb-sys in gemspec

jbourassa:
See this comment for details: https://github.com/bytecodealliance/wasmtime-rb/pull/408#issuecomment-2585983812

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1371,18 +1371,18 @@ dependencies = [
 
 [[package]]
 name = "rb-sys"
-version = "0.9.106"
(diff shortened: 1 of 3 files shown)


saulecabrera (approved):

saulecabrera:
Let's land this one and I'll try the release job on main after the fact.

jbourassa:
I also bumped the rb-sys crate to avoid version mismatch warning, and will merge once CI completes. But that alone won't solve our issue unfortunately, but won't hurt either.

Issue #418, opened 2025-01-13
No precompiled gem for Ruby 3.3


jbourassa:
Gem precompilation is broken for Ruby 3.3 which is blocking release. To fix it, we need to set the ruby version to 3.3.5 (not 3.3.0), see https://github.com/oxidize-rb/rb-sys/pull/475.

We're on an deprecated and archived version of the cross-gem action, which [explicitly requires that version match `[0-9].[0-9].0`](https://github.com/oxidize-rb/cross-gem-action/blob/12383ba9a79416b2658a9ba88f9ca7b02c599e86/src/input.ts#L67-L72). The right path forward is to upgrade to the [currently-supported cross-gem action](https://github.com/oxidize-rb/actions/tree/main/cross-gem).

When upgrading, the rustc process dies without any output:

```
/usr/local/cargo/bin/cargo rustc --crate-type cdylib --package wasmtime-rb --manifest-path /home/runner/work/wasmtime-rb/wasmtime-rb/ext/Cargo.toml --target aarch64-unknown-linux-gnu --target-dir /home/runner/work/wasmtime-rb/wasmtime-rb/tmp/rb-sys-dock/aarch64-linux/target --lib --profile release -- -C linker=aarch64-linux-gnu-gcc -L native=/usr/local/rake-compiler/ruby/aarch64-linux-gnu/ruby-3.1.0/lib -C link-arg=-pipe -C link-arg=-lm
   Compiling proc-macro2 v1.0.92
<snip...>
   Compiling postcard v1.1.1
Error: Process completed with exit code 1.
```

See [this example run](https://github.com/jbourassa/wasmtime-rb/actions/runs/12738502464/job/35501447372).

I can reproduce by SSHing into the GitHub Action worker and:
* editing the locally-installed gem source so that [`interactive?`](https://github.com/oxidize-rb/rb-sys/blob/241a1425681a96412002ab9cd992ea3e05e80585/gem/exe/rb-sys-dock#L341-L343) is false
* `rm -rf tmp` to ensure a fresh build
* `rb-sys-dock --platform aarch64-linux --directory . --ruby-versions 3.1,3.2,3.3,3.4 --build`

Interestingly, interactive or subsequent builds do not suffer from this.

[...]

PR #419, opened 2025-01-13
chore: Fix release workflow

saulecabrera:
This commit attempts to address
https://github.com/bytecodealliance/wasmtime-rb/issues/418, which is currently blocking release. This change pretty much includes all the suggestions in the linked issue.

Additionally, this change makes use of @actions/download-artifact@v4. Given that the new cross-gem action gives us control over how to upload the gem artifacts, download-artifact should now be compatible. It's also worth noting that download-artifact@v3 was deprecated on Nov 30th 2024 https://github.com/actions/download-artifact?tab=readme-ov-file#actionsdownload-artifact

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,17 +42,22 @@ jobs:
 
       - uses: oxidize-rb/actions/setup-ruby-and-rust@v1
         with:
-          ruby-version: "3.3"
+          ruby-version: "3.3.5"
           bundler-cache: true
           cargo-cache: true
           cargo-vendor: true
           cache-version: v1-${{ matrix.ruby-platform }}
 
-      - uses: oxidize-rb/cross-gem-action@main
+      - uses: oxidize-rb/actions/cross-gem@v1
+        id: cross-gem
         with:
-          version: latest
           platform: ${{ matrix.ruby-platform }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -42,18 +42,23 @@ jobs:
 
       - uses: oxidize-rb/actions/setup-ruby-and-rust@v1
         with:
-          ruby-version: "3.3"
+          ruby-version: "3.3.5"
```
jbourassa, line 45: I don't think this should be necessary, but I don't know if it hurst either.

jbourassa (commented):
I tried roughly this and can't get it to work[1], for some reason the process dies when compiling. That's what I was trying to describe in #418.

Do we have a reason to believe this version wouldn't run into the same issue?

[1]: https://github.com/jbourassa/wasmtime-rb/commit/18ccea68165f05992a7480ed8065c897f637bcef

saulecabrera:
> I tried roughly this and can't get it to work[1](https://github.com/jbourassa/wasmtime-rb/commit/18ccea68165f05992a7480ed8065c897f637bcef), for some reason the process dies when compiling. That's what I was trying to describe in #418.
> 
> Do we have a reason to believe this version wouldn't run into the same issue?

Oh I totally missed that, in that case, I don't think there's anything fundamentally different in this version that would cause it to succeed.

PR #420, opened 2025-01-14
Update cross-gem action

### LINKED ISSUE #418: No precompiled gem for Ruby 3.3

jbourassa:
Update cross-gem action, should fix #418.

Changes:
- Update the cross-gem action the release and build-gem workflows
- Update to latest upload+download artifacts
- Replace yet another Ruby 3.3 by Ruby 3.4
- Update rb-sys to fix the compilation getting killed (issue described in #418)

This _should_ work; I ran a similar workflow that included https://github.com/oxidize-rb/rb-sys/pull/487 and all the *.gem were generated, in the right folder, and had precompiled binaries for all Ruby versions ([code](https://github.com/jbourassa/wasmtime-rb/blob/3b2c7977592d068a8333058e17d128d057bb71f7/.github/workflows/release.yml), [run](https://github.com/jbourassa/wasmtime-rb/actions/runs/12756437467/job/35555185886))

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -40,28 +40,25 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: oxidize-rb/actions/setup-ruby-and-rust@v1
+      - uses: ruby/setup-ruby@v1
         with:
           ruby-version: "3.4"
-          bundler-cache: true
-          cargo-cache: true
-          cargo-vendor: true
-          cache-version: v2-${{ matrix.ruby-platform }}
 
-      - name: Set vars
-        id: vars
-        run: |
-          echo "rb-sys-version=$(bundle exec ruby -rrb_sys -e 'puts RbSys::VERSION')" >> $GITHUB_OUTPUT
-
-      - uses: oxidize-rb/cross-gem-action@main
+      - uses: oxidize-rb/actions/cross-gem@v1
+        id: cross-gem
         with:
-          version: ${{ steps.vars.outputs.rb-sys-version }}
           platform: ${{ matrix.ruby-platform }}
-          ruby-versions: ${{ join(fromJSON(needs.ci-data.outputs.result).stable-ruby-versions, ', ') }}
+          ruby-versions: ${{ join(fromJSON(needs.ci-data.outputs.result).stable-ruby-versions, ',') }}
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: cross-gem-${{ matrix.ruby-platform }}
+          path: ${{ steps.cross-gem.outputs.gem-path }}
+          if-no-files-found: error
 
       - name: Smoke gem install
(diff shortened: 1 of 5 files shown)


jbourassa:
There's 3 pending checks because the ruby version is now 3.3.5 in the matrix. Once it merges we can go update the merge requirements.

IIUC soon the matrix will include minor versions for all ruby versions, syncing between the GitHub config and ruby minor releases will create toil. I wonder if we could avoid it somehow.

saulecabrera (approved):
I'll close https://github.com/bytecodealliance/wasmtime-rb/pull/419, in favor of this one, which is a more complete state.

PR #440, opened 2025-03-03
Bump wasmtime to v30.0.2

saulecabrera:
Grabbing the latest patch release since it contains the latest fixes.

This change also mirrors the `async_stack_zeroing` change, introduced in https://github.com/bytecodealliance/wasmtime/pull/10027


ext/src/ruby_api/pooling_allocation_config.rs
```diff
@@ -100,16 +100,6 @@ impl PoolingAllocationConfig {
         Ok(rb_self)
     }
 
-    /// @def async_stack_zeroing=(enable)
-    /// @param enable [Boolean]
```
jbourassa, line 104: Let's move the documentation to `Engine.new` (engine.rs)?

jbourassa (approved):
LGTM.

The docs failure may be related to the nightly version we're pinning to. You should be able to use a more recent nightly by also bumping the yard-rustdoc gem (the format changed again but I fixed it a few weeks ago).

ext/src/ruby_api/pooling_allocation_config.rs
saulecabrera: Good catch, documented.

jbourassa:
Not sure why --target-dir stopped working, but you could try with [-o](https://doc.rust-lang.org/rustdoc/command-line-arguments.html#-o--out-dir-output-directory-path).

saulecabrera:
> Not sure why --target-dir stopped working, but you could try with [-o](https://doc.rust-lang.org/rustdoc/command-line-arguments.html#-o--out-dir-output-directory-path). I'd try it myself but won't be around the computer for a while.

Yeah I tried with `-o`, but then I get 

    Unrecognized option p


But I think something else is going on here, when I run the the (existing) command in the terminal, everything works as expected.

saulecabrera:
For some reason this line

    nightly = File.readlines("NIGHTLY_VERSION").first

was introducing a newline, and when interpolated was causing the string to split and therefore the command to be treated erroneously.

jbourassa:
Oh yeah that makes sense, the diff shows the newly added new line at end of file. But stripping it like you did is perfect.

PR #447, opened 2025-04-08
Update to `wasmtime` v31

saulecabrera:


jbourassa (approved):

saulecabrera:
Hmm unsure why the `latest` runners seem to be blocked.

saulecabrera:
Ah, it's because https://cache.ruby-lang.org/pub/misc/ci_versions/cruby.json (used by oxidize-rb/actions), no longer contains 3.1; I've removed that check form the branch protection rules.

PR #448, opened 2025-04-17
Use wasmtime-wasi APIs instead of wasi-common

jeffcharles:
This is not ready, it will not compile.

This is my work in progress for switching `wasmtime-rb` to use `wasmtime-wasi` APIs instead of `wasi-common`. This will involve having to make breaking API changes because the `wasmtime-wasi` APIs are quite different from `wasi-common`'s. Particularly around not being able to change the `stdio` streams after creating a `WasiP1Ctx`.

### DIFF
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -19,6 +19,8 @@ ruby-api = []
 winch = ["wasmtime/winch"]
 
 [dependencies]
+async-trait = "*"
+bytes = "*"
 lazy_static = "1.5.0"
 magnus = { version = "0.7", features = ["rb-sys"] }
 rb-sys = { version = "*", default-features = false, features = [
@@ -40,7 +42,7 @@ async-timer = { version = "1.0.0-beta.15", features = [
 ], optional = true }
 static_assertions = "1.1.0"
 wasmtime-environ = "=31.0.0"
-deterministic-wasi-ctx = "=0.1.34"
+deterministic-wasi-ctx = "=1.0.0"
 
 [build-dependencies]
 rb-sys-env = "0.2.2"
--- a/ext/src/helpers/file_stdin_stream.rs
+++ b/ext/src/helpers/file_stdin_stream.rs
@@ -1,0 +1,25 @@
+use std::sync::Arc;
+
+use wasmtime_wasi::{FileInputStream, StdinStream};
+
+pub struct FileStdinStream {
+    file: Arc<File>,
+}
+
+impl FileStdinStream {
+    pub fn new(file: File) -> Self {
+        Self {
+            file: Arc::new(file),
+        }
+    }
+}
+
+impl StdinStream for FileStdinStream {
+    fn stream(&self) -> Box<dyn wasmtime_wasi::InputStream> {
+        Box::new(FileInputStream::new(Arc::clone(&self.file), 0))
+    }
+
+    fn isatty(&self) -> bool {
+        false
+    }
+}
--- a/ext/src/helpers/output_limited_buffer.rs
+++ b/ext/src/helpers/output_limited_buffer.rs
@@ -1,9 +1,63 @@
+use bytes::Bytes;
 use magnus::{
     value::{InnerValue, Opaque, ReprValue},
     RString, Ruby,
 };
-use std::io;
-use std::io::ErrorKind;
+use std::io::Write;
+use std::sync::{Arc, Mutex};
+use wasmtime_wasi::{OutputStream, Pollable, StdoutStream, StreamError, StreamResult};
+
+pub struct WritePipe {
+    inner: Arc<Mutex<OutputLimitedBuffer>>,
+}
+
+impl WritePipe {
+    pub fn new(inner: OutputLimitedBuffer) -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(inner)),
+        }
+    }
+}
+
+impl Clone for WritePipe {
+    fn clone(&self) -> Self {
+        Self {
+            inner: Arc::clone(&self.inner),
+        }
+    }
+}
+
+impl StdoutStream for WritePipe {
+    fn stream(&self) -> Box<dyn wasmtime_wasi::OutputStream> {
+        let cloned = self.clone();
+        Box::new(cloned)
+    }
+
+    fn isatty(&self) -> bool {
+        false
+    }
+}
+
+#[async_trait::async_trait]
+impl Pollable for WritePipe {
+    async fn ready(&mut self) {}
+}
+
+impl OutputStream for WritePipe {
+    fn write(&mut self, bytes: Bytes) -> StreamResult<()> {
+        let mut stream = self.inner.lock().expect("Should be only writer");
+        stream.write(&bytes)
+    }
+
+    fn flush(&mut self) -> StreamResult<()> {
+        Ok(())
+    }
+
+    fn check_write(&mut self) -> StreamResult<usize> {
+        let mut stream = self.inner.lock().expect("Should be only writer");
+        stream.check_write()
+    }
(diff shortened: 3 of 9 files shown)


ext/src/helpers/file_stdin_stream.rs
```diff
@@ -0,0 +1,25 @@
...
+
+impl FileStdinStream {
+    pub fn new(file: File) -> Self {
+        Self {
+            file: Arc::new(file),
+        }
+    }
+}
+
+impl StdinStream for FileStdinStream {
```
jeffcharles, line 17: The `wasmtime-wasi` API requires types passed to `stdin` to implement `StdinStream` and I couldn't find an existing type that implementing this trait for file input.

ext/src/helpers/file_stdin_stream.rs
```diff
@@ -0,0 +1,25 @@
+use std::sync::Arc;
+
+use wasmtime_wasi::{FileInputStream, StdinStream};
+
+pub struct FileStdinStream {
+    file: Arc<File>,
```
jeffcharles, line 6: I don't actually know what fields this struct _should_ have. I tried making a similar implementation to how `OutputFile` works but the types used by the constructor functions for are different.

ext/src/helpers/output_limited_buffer.rs
```diff
@@ -21,8 +75,12 @@ impl OutputLimitedBuffer {
     }
 }
 
-impl io::Write for OutputLimitedBuffer {
-    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+impl OutputLimitedBuffer {
```
jeffcharles, line 78: We still have to use a custom type since the default in-memory implementation will trap when its capacity is exceeded and it looks like we don't want to trap when the capacity for our implementation is exceeded

ext/src/helpers/output_limited_buffer.rs
```diff
@@ -1,9 +1,63 @@
...
     value::{InnerValue, Opaque, ReprValue},
     RString, Ruby,
 };
-use std::io;
-use std::io::ErrorKind;
+use std::io::Write;
+use std::sync::{Arc, Mutex};
+use wasmtime_wasi::{OutputStream, Pollable, StdoutStream, StreamError, StreamResult};
+
+pub struct WritePipe {
```
jeffcharles, line 10: We need a custom type so we can implement `StdoutStream` which effectively requires wrapping all fields in `OutputLimitedBuffer` in the same `Arc`.

ext/src/helpers/output_limited_buffer.rs
```diff
@@ -21,8 +75,12 @@ impl OutputLimitedBuffer {
     }
 }
 
-impl io::Write for OutputLimitedBuffer {
-    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+impl OutputLimitedBuffer {
+    fn check_write(&mut self) -> StreamResult<usize> {
+        Ok(usize::MAX)
```
jeffcharles, line 80: The caller will invoke `check_write` and then invoke `write` with a buffer that fits in the `usize` returned. If it repeatedly returns 0, then the caller will hang indefinitely so this shouldn't return 0.

ext/src/ruby_api/linker.rs
```diff
@@ -54,8 +54,10 @@ impl Linker {
 
         let mut inner: LinkerImpl<StoreData> = LinkerImpl::new(engine.get());
         if wasi {
-            wasi_common::sync::add_to_linker(&mut inner, |s| s.wasi_ctx_mut())
-                .map_err(|e| error!("{}", e))?
+            wasmtime_wasi::preview0::add_to_linker_sync(&mut inner, |s| s.wasi_ctx_mut())
```
jeffcharles, line 57: The old implementation added both preview 0 and preview 1 to the linker so copied that here. There's an argument to be made this should only add preview 1 and not preview 0.

ext/src/ruby_api/wasi_ctx.rs
```diff
@@ -29,96 +25,6 @@ pub struct WasiCtx {
 type RbSelf = Obj<WasiCtx>;
 
 impl WasiCtx {
-    /// @yard
-    /// Create a new deterministic {WasiCtx}. See https://github.com/Shopify/deterministic-wasi-ctx for more details
-    /// @return [WasiCtx]
-    pub fn deterministic() -> Self {
-        Self {
-            inner: RefCell::new(wasi_deterministic_ctx()),
```
jeffcharles, line 37: There is no equivalent for this call with `wasmtime-wasi`. Instead there's a method that accepts a `WasiCtxBuilder` as input.

ext/src/ruby_api/wasi_ctx.rs
```diff
@@ -29,96 +25,6 @@ pub struct WasiCtx {
...
-        }
-    }
-
-    /// @yard
-    /// Set stdin to read from the specified file.
-    /// @def set_stdin_file(path)
-    /// @param path [String] The path of the file to read from.
-    /// @return [WasiCtxBuilder] +self+
-    fn set_stdin_file(rb_self: RbSelf, path: RString) -> RbSelf {
```
jeffcharles, line 46: The rest of the deleted calls have no equivalent calls on `WasiP1Ctx`.

ext/src/ruby_api/wasi_ctx_builder.rs
```diff
@@ -222,20 +225,32 @@ impl WasiCtxBuilder {
         rb_self
     }
 
+    /// @yard
+    /// Make the context deterministic. See See https://github.com/Shopify/deterministic-wasi-ctx for more details
+    /// @def add_determinism()
+    /// @return [WasiCtxBuilder] +self+
+    pub fn add_determinism(rb_self: RbSelf) -> RbSelf {
```
jeffcharles, line 232: Note that we still need to call `replace_scheduling_functions` (and the preview 0 equivalent) with `deterministic_wasi_ctx` passing the linker as an argument if there's a desire to prevent modules from being able to sleep.

ext/src/ruby_api/wasi_ctx_builder.rs
```diff
@@ -222,20 +225,32 @@ impl WasiCtxBuilder {
...
 
         if let Some(stdin) = inner.stdin.as_ref() {
             match stdin {
                 ReadStream::Inherit => builder.inherit_stdin(),
-                ReadStream::Path(path) => {
-                    builder.stdin(file_r(ruby.get_inner(*path)).map(wasi_file)?)
-                }
+                ReadStream::Path(path) => builder.stdin(
+                    file_r(ruby.get_inner(*path))
+                        .map(|path| FileStdinStream::new(FileInputStream::new(&path, 0)))?,
```
jeffcharles, line 247: I cannot figure out how to construct `FileInputStream`. The link for the `&File` type in https://docs.rs/wasmtime-wasi/31.0.0/wasmtime_wasi/struct.FileInputStream.html#method.new doesn't exist. Looking at the underlying code, the `File` type used does not appear to be exported by `wasmtime-wasi`. We may want to write our own equivalent of `FileInputStream`.

ext/Cargo.toml
```diff
@@ -19,6 +19,8 @@ ruby-api = []
 winch = ["wasmtime/winch"]
 
 [dependencies]
+async-trait = "*"
+bytes = "*"
```
jeffcharles, line 23: These two are required by traits we have to implement.

jeffcharles:
There's an additional compilation failure with `WasiCtx` where `get_inner` fails because `WasiP1Ctx` does not appear to implement `Clone`. I'm not really sure how to handle that right now.

PR #458, opened 2025-05-22
Do not merge - Debugging Windows build failures

jeffcharles:


jeffcharles:
Looks like the `sysroot` changed from `C:\msys64` to `C:\hostedtoolcache\windows\Ruby\3.2.8\x64\msys64` but the `setup-rust-and-ruby` action assumes the sysroot is `C:\msys64` so Clang doesn't provided with the correct path to find the header files it needs.

PR #480, opened 2025-07-02
Add rust-toolchain.toml to address failing build native gems

### LINKED ISSUE #479: Build native gems workflow failing
See https://github.com/bytecodealliance/wasmtime-rb/actions/runs/16024387186/job/45208682436

When cross-compiling, there's a failure happening due to an apparent incompatibility between the cargo version used and the Rust edition declared in our extension (2024). 

According to my investigation, it seems that `rb-sys-dock build ..`  [is defaulting to use the docker-provided Cargo/Rust version (1.84)](https://github.com/oxidize-rb/rb-sys/blob/main/gem/exe/rb-sys-dock#L238) instead of using the stable version (1.88 at the time of writing).

I believe the solution here is to ensure that we force the usage of the `stable` toolchain, like we do in all the other GitHub workflows. I'm wondering if the only reaslistic and future-proof solution is to force the installation of the stable toolchain in `rb-sys-dock` every time instead of relying on the one provided by Docker. 

cc/ @ianks I'm interested on your thoughts.

jeffcharles:
I think specifying this file should cause the current stable toolchain to be used which could fix #479. See https://github.com/jeffcharles/wasmtime-rb/actions/runs/16029783762/job/45227084441#step:4:447 for what this did on my fork of the repo for the _Build native gems_ workflow where it's failing on main.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "stable"
+components = ["clippy", "rustfmt"]
+targets = ["wasm32-wasip1"]
+profile = "default"


saulecabrera:
IIRC, we were very intentional about not introducing a `rust-toolchain.toml` to ensure that we can delegate the Rust/Ruby setup to `oxidize-rb` entirely, which is what ultimately orchestrates the builds.

Are the failures here related to the introduction of `rust-toolchain.toml`?

PR #503, opened 2025-09-29
Integrate direnv with flakes

saulecabrera:
This change makes local development easier for NixOS-based systems by introducing a flake which declares the Ruby dependencies of this project. This is useful in cases in which we don't want (or need) to declare Ruby and it's dependencies globally.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,16 @@
+ {
+    inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    inputs.flake-utils.url = "github:numtide/flake-utils";
+
+    outputs = { nixpkgs, flake-utils, ... }:
+      flake-utils.lib.eachDefaultSystem (system: {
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,16 @@
+ {
+    inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    inputs.flake-utils.url = "github:numtide/flake-utils";
+
+    outputs = { nixpkgs, flake-utils, ... }:
+      flake-utils.lib.eachDefaultSystem (system: {
+        devShells.default = nixpkgs.legacyPackages.${system}.mkShell {
+          packages = with nixpkgs.legacyPackages.${system}; [
```
saulecabrera, line 8: Note that I could declare Rust dependencies as well, however, I'm opting not to since: (i) I have Rust globally installed in my system (ii) this is pretty much a convenience for local development for myself mostly (iii) we can tune the flake definition in a follow-up if needed.

PR #588, opened 2026-04-22
update to Wasmtime 44.0.0

jacobsteves:


ext/src/ruby_api/linker.rs
```diff
@@ -171,8 +171,8 @@ impl Linker {
 
         let ruby = Ruby::get_with(store);
         match ext {
-            None => Ok(None),
-            Some(ext) => ext.wrap_wasmtime_type(&ruby, store.into()).map(Some),
+            Err(_) => Ok(None),
```
saulecabrera: I _think_ this is unsound and it would be best to surface the error instead. Wasmtime 44 gracefully handles OOM, and the public APIs reflect that change, like in this case returning an error rather than simply returning `Option<T>`. 

See https://github.com/bytecodealliance/wasmtime/pull/12858/

saulecabrera: Unsure how this relates to the rest of the API surfaced by wasmtime-rb though, I think we might need to audit it.

saulecabrera (commented):
Looking good, left one comment that I think we need to solve before merging this.

ext/src/ruby_api/linker.rs
jeffcharles: FWIW, the C API for the linker has [a similar change](https://github.com/bytecodealliance/wasmtime/pull/12858/changes#diff-1f3631bde88b3cbd21972da074c69919911b310fd311e8afd55d44662ecdd5c8). At the moment, the Rust API returns an error saying the item doesn't exist. Is the worry that the Rust API might change in the future to return an error indicating an allocation failure as an additional failure path?

saulecabrera: > Is the worry that the Rust API might change in the future to return an error indicating an allocation failure as an additional failure path?

It already does FWIW:

```rust
    pub fn get(
        &self,
        mut store: impl AsContextMut<Data = T>,
        module: &str,
        name: &str,
    ) -> Result<Extern>
    where
        T: 'static,
    {
        let store = store.as_context_mut().0;
        match self._get(module, name) {
            // Safety: `T` is connecting the linker and store.
            Some(def) => Ok(unsafe { def.to_extern(store)? }),
            None => bail!("missing definition for `{module}::{name}`"),
        }
    }

```

If you follow the path through `def.to_extern`, it might error with `Result<Extern, OutOfMemory>`: 

```rust
    pub(crate) unsafe fn to_extern(&self, store: &mut StoreOpaque) -> Result<Extern, OutOfMemory> {
        match self {
            Definition::Extern(e, _) => Ok(e.clone()),
            // SAFETY: the contract of this function is the same as what's
            // required of `to_func`, that `T` of the store matches the `T` of
            // this original definition.
            Definition::HostFunc(func) => unsafe { Ok(func.to_func(store)?.into()) },
        }
    }
```

jacobsteves: Done - I'm now surfacing the errors. It is a breaking API change but its aligned with the wasmtime API and will be a major bump anyway so I think thats reasonable

saulecabrera (approved):
Thanks!

PR #597, opened 2026-05-08
Component host functions

### LINKED ISSUE #433: Component host functions
Add host functions for components.

Wasmtime has 2 APIs for adding host functions to components:
1. The typed API (e.g. [`func_wrap`](https://docs.wasmtime.dev/api/wasmtime/component/struct.LinkerInstance.html#method.func_wrap)) which requires the signature be known at compile time.
2. The dynamic API (e.g. [`func_new`](https://docs.wasmtime.dev/api/wasmtime/component/struct.LinkerInstance.html#method.func_new)), where the params are in a `&[Val]` and return values are to be written to a `&mut[Val]`.

I think (1) essentially requires a rust compiler thus out of question. We're left with (2). But how do we wrap such API in Ruby?

### Option A: follow the crate's API

Params are Ruby values converted from Wasm, return values are tagged with their type:

```ruby
# Imagine the following WIT definitions:

# f(param1: string) -> result<a>
#
# record a {
#   field1: u32,
#   field2: string,
# }

T = Wasmtime::Component::Type # Hypothetical API, does not exist yet

linker_istance.func_new("f") do |_store, param1|
  T::Result.wrap(
    Wasmtime::Component::Result.ok(
      T::Record.wrap(
        field1: T::U32.wrap(42),
        field2: T::String.wrap("foo"),
      )
    )
  )
end
```

With this approach, params are not type-checked. E.g. `param1` could be of any type; whatever the caller decides to send. The host function could implement some limited type checking (limited because it can't distinguish between a u8 or u32, for example).

### Option B: type the Function
Similar to what's done for core Wasm functions, specify the return type when calling `#func_new`:
```ruby
T = Wasmtime::Component::Type # Hypothetical API, does not exist yet

A = T::Record.new(
  "field1" => T::U32
  "field1" => T::String
)
ReturnType = T::Result.new(A)

linker_istance.func_new("f", ReturnType) do |_store, param1|
  record = { "field1" => 42, "field2" => "foo" }
  Wasmtime::Component::Result.ok(record)
end
```

This approach can be extended to params to support params type checking (possibly with a performance hit).

We might be able to generate those types directly by parsing a WIT file. 

----

Both options A and B require mapping a Ruby object to a Wasm component type definition in a *standalone* way. By standalone I mean not using an index to point into a type in a component, i.e. not using the [`wasmtime::component::Type`](https://docs.wasmtime.dev/api/wasmtime/component/enum.Type.html) enum. The implication of this is the [convert code](https://github.com/bytecodealliance/wasmtime-rb/blob/668d461d34b083e8a4c30647f6c4525e093262f9/ext/src/ruby_api/component/convert.rs#L99-L284) needs to be duplicated or a new abstraction introduced.

There may be a better way. If you think of one, please share! Unless we find a better way, I suggest waiting until users really need this feature before implementing this.

william-stacken:
Fixes #433 

This PR introduces host functions for components, with strict type validation such that if the function signatures do not match between what the component expects and the host function, an error would be raised. Let me know what you think about the direction!


ext/src/ruby_api/component/linker.rs
```diff
@@ -254,6 +506,108 @@ impl<'a> LinkerInstance<'a> {
...
+                        return Err(wasmtime::Error::msg(format!(
+                            "expected 1 result, got {}",
+                            result_array.len()
+                        )));
+                    }
+                    unsafe { result_array.as_slice()[0] }
+                } else {
+                    proc_result
+                };
+
```
saulecabrera: Can you add a bit more context on why this approach is needed/beneficial? What if the return type is a single list/tuple itself? e.g., 

```wit
import get-numbers: func() -> list<s32>;
```

I think that the natural implementation from Ruby would be something like:

```ruby
linker.root do |root|
  root.func_new("get-numbers", [], [t.list(t.s32)]) do
    [1, 2, 3]
  end
end
```

But if I am understanding correctly, this snippet will flow though the `1` arm and error out?

With this approach if we actually want to return the array, we would need to double wrap:

```ruby
linker.root do |root|
  root.func_new("get-numbers", [], [t.list(t.s32)]) do
    [[1, 2, 3]]
  end
end
```
Related: I think we are missing more test cases in which the components return container types e.g., list, tuple.

ext/src/ruby_api/component/linker.rs
```diff
@@ -10,16 +12,34 @@ use crate::{
...
+    Value,
 };
-use wasmtime::component::{Linker as LinkerImpl, LinkerInstance as LinkerInstanceImpl};
+use wasmtime::component::{Linker as LinkerImpl, LinkerInstance as LinkerInstanceImpl, Val};
 use wasmtime_wasi::{ResourceTable, WasiCtx};
 
+/// Stores type information for a registered host function
+#[derive(Clone, Debug)]
+struct FuncTypeInfo {
+    #[allow(dead_code)]
```
saulecabrera: I think we can get rid of this?

ext/src/ruby_api/component/linker.rs
```diff
@@ -140,6 +171,141 @@ impl Linker {
             .map_err(|e| error!("{}", e))
     }
 
+    /// Validate that host functions defined via func_new match the component's expected import signatures
+    fn validate_host_function_signatures(
```
saulecabrera: We have multiple helper functions defined inside this function, could we make a different module with all this logic instead? It might make it easier to reason about.

saulecabrera:
I am taking a look at this, it might take a while, since this is a considerable chunk of code to go through.

jbourassa:
> Let me know what you think about the direction!

Not really feedback, but a thought: [wasmex](https://github.com/tessi/wasmex) uses the wit file when defining a component, instead of having a DSL for defining signatures in Elixir. [Documented here](https://github.com/tessi/wasmex/blob/65b81ab3bd5d2a56ba8ee42b11d6dfe5c710ff7c/lib/wasmex/components/component_server.ex#L1-L97).

I think it's worth considering, but haven't given it much thought recently though; maybe there's a valid reason not to go with that approach. Feel free to push back!

spec/unit/component/linker_spec.rb
```diff
@@ -35,6 +35,440 @@ module Component
...
+        end
+
+        context "complex types" do
+          it "defines a function with record types" do
+            point_type = t.record("x" => t.s32, "y" => t.s32)
+
+            linker.root do |root|
+              root.func_new("make-point", [t.s32, t.s32], [point_type]) do |x, y|
+                {"x" => x, "y" => y}
+              end
```
saulecabrera: In `validate_host_function_signatures` we are structurally validating that records match i.e., names and types must be identical and in the same order; we should probably add a test to ensure that such branch is covered?

ext/src/ruby_api/component/linker.rs
```diff
@@ -140,6 +171,141 @@ impl Linker {
...
+                            export_name.to_string()
+                        } else {
+                            format!("{}/{}", path, export_name)
+                        };
+                        validate_imports_recursive(&nested_path, &export_item, func_types, engine)?;
+                    }
+                }
+                _ => {
+                    // Other types (Module, CoreFunc, Type, etc.) are not validated here
+                }
```
saulecabrera: Meaning that this validation is handed over to wasmtime?

saulecabrera (commented):
In general the approach looks reasonable, however, I am concerned about the maintainability and performance aspects of the type system: we are in a sense duplicating some of the type checks that normally should be handed to Wasmtime.

To say: I am not fully convinced that the UX gains justify the complexity and the potential performance hit.

I think that a reasonable alternative to explore is a variant of Option A, as described in https://github.com/bytecodealliance/wasmtime-rb/issues/433, which is somewhat similar to [what `wasmtime-py` does](https://github.com/bytecodealliance/wasmtime-py/blob/main/wasmtime/component/_linker.py#L229) (IIUC) lazy value coercion checks rather than eager type checks. Trade-offs that I can think of: (i) errors will surface at function call time rather than instantiation time (ii) we lose the homogeneous nature of the Component/Core APIs. 

My assumption is that it shouldn't be too hard to try out the Option A variant on top of the changes present here, IIUC, it will be mostly a matter of removing the type check that is happening at instantiation time. Having some performance numbers will probably help us finalize the decision. 

I looked a bit at Wasmex's approach, and my understanding is that their API is philosophically different to ours: they do not try to stay as close to Wasmtime's API, which is why I think the approach of using a WIT file suits their needs best. Personally I was not able to convince myself that we could adopt a similar approach.

ext/src/ruby_api/component/linker.rs
william-stacken: Good point. I hve now made changes to type the return values, so that it is possible to distinguish between an array and multiple return values.

william-stacken:
I have now removed the instantiation-time type validations. I also added a "type wrapper" class for return values of the host functions as described in Option A of the issue, which allows wasmtime to perform type checks. With this approach, the host functions can perform type validations and raise an exception if they wish.

spec/unit/component/linker_spec.rb
saulecabrera, line 1: Can we add some tests covering round trips for the following: s8, u8, s16, u16, s64, u64, f32, f64, char, variant, enum and flags? I think they are not covered in the tests.

spec/unit/component/linker_spec.rb
```diff
@@ -35,6 +35,442 @@ module Component
             .to be_instance_of(Wasmtime::Component::Instance)
         end
       end
+
+      describe "LinkerInstance#func_new" do
+        let(:t) { Type }
+
+        context "simple host functions" do
+          it "defines a function with primitives" do
```
saulecabrera: With the new implementation (validation at call time, rather than at definition time), we would need to call the function in order to validate the right behavior, right?

ext/src/ruby_api/component/linker.rs
```diff
@@ -254,6 +351,114 @@ impl<'a> LinkerInstance<'a> {
...
+                let converted =
+                    convert::wrapped_to_component_val(result_value, None).map_err(|e| {
+                        // Store type errors on StoreData as well
+                        store_context.data_mut().set_error(e);
+                        wasmtime::Error::msg("")
+                    })?;
+                results[0] = converted;
+                Ok(())
+            }
+            n => {
```
saulecabrera: Should we have a test for this branch? I think we currently only cover `<tuple<t>>`, which still counts as a single value?

spec/unit/component/linker_spec.rb
saulecabrera, line 1: Maybe I missed it, but I do not see tests for arity mismatch? If so, could we add some of those as well?

ext/src/ruby_api/component/linker.rs
```diff
@@ -254,6 +351,114 @@ impl<'a> LinkerInstance<'a> {
...
+    wasmtime::StoreContextMut<'_, StoreData>,
+    wasmtime::component::types::ComponentFunc,
+    &[Val],
+    &mut [Val],
+) -> wasmtime::Result<()>
+       + Send
+       + Sync
+       + 'static {
+    move |mut store_context: wasmtime::StoreContextMut<'_, StoreData>,
+          _func: wasmtime::component::types::ComponentFunc,
```
saulecabrera: Given that all the validation is happening at this point, I am wondering what do we gain with the `t.wrap` approach. Couldn't we drop it entirely and rely on the type information present in `wasmtime::component::types::ComponentFunc` for validation when calling `wrapped_to_component_val`?

ext/src/ruby_api/component/linker.rs
william-stacken: Good point. Looking into this I realized that this branch might be dead code, it doesn't seem like it is possible to specify multiple return values for functions in WIT.

From [here](https://component-model.bytecodealliance.org/design/wit.html):
> To express a function that returns multiple values, you can use any compound type (such as [tuples](https://component-model.bytecodealliance.org/design/wit.html#tuple) or [records](https://component-model.bytecodealliance.org/design/wit.html#record)).

So maybe we should just raise an error if a wasm component returns multiple values and have it be unsupported for now.

spec/unit/component/linker_spec.rb
william-stacken: True. I think this is already covered by the integration tests, so I will reduce the number of non-integration tests to the bare minimum (testing that providing a block works as expected and that providing no block fails)

spec/unit/component/linker_spec.rb
william-stacken: Do you mean that wasmtime-rb should raise an error if the arity of the block does not match? One thing that could complicate this is if the block has splats or keyword arguments, would we disallow such blocks entirely and enforce positional block arguments only?

ext/src/ruby_api/component/linker.rs
william-stacken: Yes, I think you are right. Removing the `t.wrap` approach even means we can rely on the existing `component_val_to_rb` instead of the new `wrapped_to_component_val`.

The one caveat is that `component_val_to_rb` needs a `StoreContextValue` which is currently unused, and this is not available in the component function closure. I changed it into acceping an `Option` instead, so that it is possible to supply `None`. I assume that the `StoreContextValue` is intended for use in the future? Does it make sense to switch to an `Option`?

spec/unit/component/linker_spec.rb
saulecabrera: Yeah, that is a good point and one of the trade-offs of this approach I believe.

saulecabrera (commented):
I think this is looking good, thanks for iterating on this. One last thing: I think we should add a tests that exercises a `void` return function invocation? I do not see one for the integration tests?

saulecabrera (approved):
Thanks!

PR #599, opened 2026-05-20
Expose `wasm_exceptions` engine option to Ruby

### LINKED ISSUE #598: `Engine.new` doesn't expose `wasm_exceptions`
`wasmtime::Config` has a `wasm_exceptions(bool)` method (gated on the default-enabled `gc` feature, so it's built into the released gem), but it isn't surfaced as a keyword on `Wasmtime::Engine.new`.
Any module that uses the [exception-handling proposal][eh] fails to even parse, with no way to opt in from Ruby.

This affects all wasm produced by clang with `-mllvm -wasm-enable-sjlj` (the standard recipe for `<setjmp.h>` in wasm).
A single `setjmp` call in the source is enough to pull in EH instructions.

[eh]: https://github.com/WebAssembly/exception-handling

## How to reproduce

```ruby
require "wasmtime"

WAT = <<~WAT
  (module
    (tag $e)
    (func (export "raise")
      throw $e))
WAT

def attempt(label, **cfg)
  Wasmtime::Module.new(Wasmtime::Engine.new(**cfg), WAT)
  puts "#{label}: OK"
rescue => e
  puts "#{label}: #{e.class}: #{e.message}"
end

attempt("without options      ")
attempt("wasm_exceptions: true", wasm_exceptions: true)
```

```console
$ ruby repro.rb
without options      : Wasmtime::Error: Could not build module: failed to parse WebAssembly module
wasm_exceptions: true: ArgumentError: Unknown option: :wasm_exceptions
```

Line 1 — engine refuses to parse the module (expected: EH is off by default).
Line 2 — no Ruby-side knob to turn it on.

## Environment

- `wasmtime` gem: 44.0.0
- Ruby: 3.4.8
- Platform: arm64-darwin (also reproducible on x86_64-linux)

takahashim:
Closes #598.

Adds a dispatch arm in `hash_to_config` so the existing `wasmtime::Config::wasm_exceptions` method becomes reachable from `Engine.new`.

## Verification

The reproducer from #598, before:

```
without options      : Wasmtime::Error: Could not build module: failed to parse WebAssembly module
wasm_exceptions: true: ArgumentError: Unknown option: :wasm_exceptions
```

After:

```
without options      : Wasmtime::Error: Could not build module: failed to parse WebAssembly module
wasm_exceptions: true: OK
```

### DIFF
--- a/ext/src/ruby_api/config.rs
+++ b/ext/src/ruby_api/config.rs
@@ -42,6 +42,7 @@ define_rb_intern!(
     POOLING => "pooling",
     ON_DEMAND => "on_demand",
     WASM_REFERENCE_TYPES => "wasm_reference_types",
+    WASM_EXCEPTIONS => "wasm_exceptions",
     ASYNC_STACK_ZEROING => "async_stack_zeroing",
 );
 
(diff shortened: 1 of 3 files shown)


jbourassa (approved):

jbourassa:
Thanks!

PR #601, opened 2026-05-23
feat: add typed memory accessors to `Wasmtime::Memory`

takahashim:
## Motivation

Currently, to retrieve an integer value from linear memory, we need to write something like `memory.read(out_p, 4).unpack1("l<")`.
This approach is verbose, difficult to understand, and prone to errors.
Therefore, we want to enable writing `memory.read_i32(out_p)` instead.

## Changes

Introduce paired read/write methods for the WebAssembly numeric types (little-endian):

- `#read_i32` / `#write_i32`
- `#read_u32` / `#write_u32`
- `#read_i64` / `#write_i64`
- `#read_u64` / `#write_u64`
- `#read_f32` / `#write_f32`
- `#read_f64` / `#write_f64`

Out-of-bounds access raises `Wasmtime::Error`, consistent with `#read` / `#write`.

### DIFF
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -227,6 +227,132 @@ impl<'a> Memory<'a> {
     }
 
     /// @yard
+    /// Read a little-endian +i32+ starting at +offset+.
+    ///
+    /// @def read_i32(offset)
+    /// @param offset [Integer]
+    /// @return [Integer]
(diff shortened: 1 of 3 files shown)


saulecabrera (approved):
For future PRs involving changes to the public API, please open an issue first for discussion. `wasmtime-rb`'s minimal API surface is intentional, we try to keep it as close as upstream as possible. In this case, the change is small enough that I think it is reasonable to land.

PR #602, opened 2026-05-25
bump rb_sys

jacobsteves:
`build-native-gems` is failing on main after the wasmtime bump because wasmtime requires rustc >= 0.93 but the step uses a docker container thats pulling an old version. Bumping the version should fix it.

`build-native-gems` doesn't run on PRs so we didn't catch it there.

Manual trigger https://github.com/bytecodealliance/wasmtime-rb/actions/runs/26407150802

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1440,18 +1440,18 @@ dependencies = [
 
 [[package]]
 name = "rb-sys"
-version = "0.9.127"
+version = "0.9.128"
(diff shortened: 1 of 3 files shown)


jeffcharles:
Can you also update the `rb-sys` crate? It'll probably still work if they're a little out-of-sync but probably best if they're not.

jeffcharles (approved):

Issue #604, opened 2026-05-30
Ruby prebuilt gems size question


omohokcoj:
Prebuilt Linux gem shared library .so builds include debug symbols, making them ~300MB. I'm curious whether the debug symbols could be stripped in gem releases, or if including them is intentional?

https://rubygems.org/gems/wasmtime/versions/45.0.0-x86_64-linux

saulecabrera:
I do not recall a dedicated discussion around keeping debug symbols. They are handy for local debugging / profiling, but I think the is room for improvement when it comes to the pre-compiled gems, e.g., have a profile that strips them from the final binary.

jbourassa:
Closed by #610

PR #610, opened 2026-06-07
Strip debug symbols from gem release shared object

omohokcoj:
Related issue: https://github.com/bytecodealliance/wasmtime-rb/issues/604 
It reduces ruby gem download size from ~190MB to ~29MB

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,5 +10,5 @@ exclude = [
 
 [profile.release]
 codegen-units = 1 # more llvm optimizations
-debug = 2         # make perfomance engineers happy
+strip = true      # strip debug symbols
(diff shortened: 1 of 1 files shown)


jbourassa:
👍 I can get behind this, folks working on the performance of the crate can still enable the debug symbols when needed.

jacobsteves:
I'm good with this too, I believe it'll actually fix https://github.com/bytecodealliance/wasmtime-rb/issues/581 as well (see  [this comment](https://github.com/bytecodealliance/wasmtime-rb/issues/581#issuecomment-4661039415))

jbourassa (approved):

PR #611, opened 2026-06-07
Add read_cstring/write_cstring memory helpers

omohokcoj:
Following the memory-access helpers added recently in https://github.com/bytecodealliance/wasmtime-rb/pull/601 could we please also have read_cstring/write_cstring - reading nul-terminated C strings is a hot path when working with C libraries, and doing the nul byte scan in Ruby is very inefficient.

### DIFF
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -383,6 +383,12 @@ impl<'a> Memory<'a> {
     /// @return [void]
     pub fn write_cstring(&self, offset: usize, value: RString) -> Result<(), Error> {
         let slice = unsafe { value.as_slice() };
+        if slice.contains(&0) {
+            return Err(Error::new(
+                Ruby::get_with(value).exception_arg_error(),
+                "string contains null byte",
+            ));
+        }
         let len = slice.len();
         let mut context = self.store.context_mut()?;
         let dst = self
--- a/ext/src/ruby_api/memory.rs
+++ b/ext/src/ruby_api/memory.rs
@@ -353,6 +353,49 @@ impl<'a> Memory<'a> {
     }
 
     /// @yard
+    /// Read a NUL-terminated C string starting at +offset+ as a UTF-8 +String+.
+    ///
+    /// @def read_cstring(offset)
+    /// @param offset [Integer]
+    /// @return [String]
+    pub fn read_cstring(ruby: &Ruby, rb_self: Obj<Self>, offset: usize) -> Result<RString, Error> {
+        let context = rb_self.store.context()?;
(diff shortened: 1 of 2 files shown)


tschneidereit:
Hi @omohokcoj since I've seen you request copilot reviews on PRs a few times now: please note that the BA has an [AI tool use policy](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md) that asks that contributors don't use automated AI tools in this way. If you want to use copilot reviews, one approach is to do so in your own repo before opening a PR upstream. Thanks!

omohokcoj:
@tschneidereit thanks for the heads up - I didn't request Copilot code review manually, but it looks like GitHub does this automatically for all pull requests without a notice/prompt. I'll have this feature disabled.

image

saulecabrera:
> but it looks like GitHub does this automatically for all pull requests without a notice/prompt. I'll have this feature disabled.

We had other PRs in the last two weeks in which no code review was automatically triggered e.g., https://github.com/bytecodealliance/wasmtime-rb/pull/601. I am not entirely sure what heuristic is used to trigger the review though. 

@tschneidereit At any rate, I noticed that the _Copilot Code Review_ setting was enabled by default in the repository. I have now disabled it. Hopefully that will reduce the noise.

tschneidereit:
I just disabled it globally for the BA. It took a support ticket and waiting a few weeks to learn how to even do that: turns out you have to enable copilot in an org to have access to the settings page that allows you to disable *use* of copilot in that same org. You've probably been able to see that setting in the repo now for the same reason ...

ext/src/ruby_api/memory.rs
```diff
@@ -352,6 +352,51 @@ impl<'a> Memory<'a> {
...
+    /// @return [void]
+    pub fn write_cstring(&self, offset: usize, value: RString) -> Result<(), Error> {
+        let slice = unsafe { value.as_slice() };
+        let len = slice.len();
+        let mut context = self.store.context_mut()?;
+        let dst = self
+            .get_wasmtime_memory()
+            .data_mut(&mut context)
+            .get_mut(offset..)
+            .and_then(|s| s.get_mut(..len + 1))
```
saulecabrera, line 398: Shouldn't this validate against arbitrary null bytes in the incoming Ruby string?

omohokcoj: @saulecabrera makes sense, validation has been added in https://github.com/bytecodealliance/wasmtime-rb/pull/611/commits/48f04704d472bd23840e354aa171248fa1bc6c50

saulecabrera (approved):

PR #613, opened 2026-06-21
Fix rb_gc_adjust_memory_usage double-counting

omohokcoj:
`Memory` wrapper tracked wasm memory via its own `ManuallyTracked`, but that memory lives in the `Store` and is already reported to Ruby by the store's `TrackingResourceLimiter` causing double-counting:

```ruby
require "wasmtime"

pages = 200
bytes = pages * 65_536
store = Wasmtime::Store.new(Wasmtime::Engine.new)

before = GC.stat(:malloc_increase_bytes)
Wasmtime::Memory.new(store, min_size: pages)
delta = GC.stat(:malloc_increase_bytes) - before

puts "linear memory: %d bytes | reported to GC: %d bytes | %.1fx\n" % [bytes, delta, delta.to_f / bytes]

# output:  linear memory: 13107200 bytes | reported to GC: 26216600 bytes | 2.0x - double-counting
```

This PR removes redundant GC stats tracking from `Memory` that double-counted wasm memory in Ruby (since the store's resource limiter already reports it).

### DIFF
--- a/ext/src/ruby_api/externals.rs
+++ b/ext/src/ruby_api/externals.rs
@@ -221,7 +221,7 @@ impl<'a> WrapWasmtimeType<'a, Extern<'a>> for wasmtime::Extern {
                 ruby.obj_wrap(Global::from_inner(store, *global)),
             )),
             wasmtime::Extern::Memory(mem) => Ok(Extern::Memory(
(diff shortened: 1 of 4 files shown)


saulecabrera (approved):
LGTM, thanks!

PR #620, opened 2026-07-02
Add Component::Instance#get_export_index/#get_resource and WIT resour…

teleyos:
…ce support

Instance previously only exposed #get_func, making it impossible to fully drive a component that exports a WIT `resource` (e.g. a constructor plus instance methods) from Ruby: there was no way to look up the resource's exported type, and Val::Resource/Type::Own/Type::Borrow conversions were hard `not_implemented!` stubs.

- Instance#get_export_index / #get_resource wrap wasmtime's own methods of the same name; the shared lookup helper now accepts a previously-resolved ExportIndex (bare or nested in an Array) so lookups can be cached and chained, matching wasmtime's own doc example, and internally prefers get_export_index over get_export since the ComponentItem it discards is never used.
- New Wasmtime::Component::Resource wraps ResourceAny + its owning Store, requiring an explicit #resource_drop (wasmtime requires every ResourceAny, even borrows, to be explicitly dropped; there's no safe GC finalizer hook for re-entering a Store during collection).
- New Wasmtime::Component::ExportIndex and ResourceType wrap the matching wasmtime types; both are inert Copy data with no Store dependency, so unlike Resource they need no GC mark.
- convert.rs now lifts/lowers resource values for the Func::invoke path (calling exported functions), and guards against passing a Resource obtained from one Store into a call on a different Store, which would otherwise silently index into the wrong store's resource table.
- Host-defined import functions (Linker#root.func_new) still don't support resources in either direction; that path never had a Ruby-visible Store handle to thread through, and stays out of scope here.

### DIFF
--- a/ext/src/ruby_api/component.rs
+++ b/ext/src/ruby_api/component.rs
@@ -1,7 +1,9 @@
 mod convert;
+mod export_index;
 mod func;
 mod instance;
 mod linker;
+mod resource;
 mod wasi_command;
 
 use super::root;
@@ -17,9 +19,11 @@ use magnus::{
 use rb_sys::tracking_allocator::ManuallyTracked;
 use wasmtime::component::Component as ComponentImpl;
 
+pub use export_index::ExportIndex;
 pub use func::Func;
 pub use instance::Instance;
 pub use linker::Linker;
+pub use resource::{Resource, ResourceType};
 pub use wasi_command::WasiCommand;
 
 pub fn component_namespace(ruby: &Ruby) -> RModule {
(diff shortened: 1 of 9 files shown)


jeffcharles:
Please review [the Bytecode Alliance's policy on AI Tool Use](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md)

teleyos:
> Please review [the Bytecode Alliance's policy on AI Tool Use](https://github.com/bytecodealliance/governance/blob/main/AI_TOOL_POLICY.md)

Thank you @jeffcharles for the guidance and I apologize to the maintainers for the bad PR, I'll work on a better one.

Issue #619, opened 2026-07-01
Wasmtime::Component::Instance.get_func(handle) -> form of handle unclear


teleyos:
Hi, i have a game.wasm component.
If i do `wasm-tools component wit game.wasm` i get 
```wit
package root:component;

world root {
  export teleyos:connect4/types@0.0.1;
}
package teleyos:connect4@0.0.1 {
  interface types {
    enum error-code {
      col-not-empty,
      out-of-bound,
    }

    type error = tuple<error-code, string>;

    type play-result = result<u8, error>;

    enum player {
      first,
      second,
    }

    type grid = list<option<player>>;

    type winner = player;

    resource state {
      constructor();
      play: func(index: s8) -> play-result;
      check-finished: func() -> tuple<bool, option<winner>>;
    }
  }
}
```
I want the name or path that i need to give to give to [get_func](https://bytecodealliance.github.io/wasmtime-rb/latest/Wasmtime/Component/Instance.html#get_func-instance_method) in order to get the check_finished [Func](https://bytecodealliance.github.io/wasmtime-rb/latest/Wasmtime/Component/Func.html).

I tried a lot of possibilities but i struggle to find which path to give because it's in root > teleyos:connect4@0.0.1 > types > state > check-finished and i don't know how to articulate that even after looking at the docs and source code.

I have this for now :
```ruby
require 'wasmtime'

engine = Wasmtime::Engine.new
component = Wasmtime::Component::Component.from_file(engine, "./game.wasm")
store = Wasmtime::Store.new(engine, {})
linker = Wasmtime::Component::Linker.new(engine)

instance = linker.instantiate(store, component)

check_finished_func = instance.get_func(["root", "teleyos:connect4@0.0.1","types","state","check-finished"])

if (check_finished_func == nil) then
  puts "didn't find function"
  puts check_finished_func
  exit
end
```

Update: Working on implementing Resource bindings and exposing a way to discover exports and their names.

saulecabrera:
I think this is expected since resources are not fully supported yet, see https://github.com/bytecodealliance/wasmtime-rb/issues/434. Can you share what is the error that you get when trying to run the snippet above? 

Edit: I assume that you are hitting `not_implemented!(ruby, "Resource not implemented")`?

PR #623, opened 2026-07-03
Feature/adding missing instance methods

teleyos:
This PR aims to enable users of the gem to use resources and to improve the get_func docs to talk about the canonical way to point to a function [ "packagename/interface@x.x.x" , "[constructor]my-resource" ].

One caveat is that the GC won't free a resource in order to avoid double freeing or freeing before intended so we have to free by hand using resource drop.

Example :
```ruby

IFACE = "somepackage/someinterface@x.x.x"

engine = Wasmtime::Engine.new

wasi = Wasmtime::WasiConfig.new
  .inherit_stdin
  .inherit_stderr
  .inherit_stdout
store = Wasmtime::Store.new(engine, wasi_config: wasi)

component = Wasmtime::Component::Component.from_file(engine, "./component_file.wasm")
linker = Wasmtime::Component::Linker.new(engine)
instance = linker.instantiate(store, component)

some_resource_constructor = instance.get_func([IFACE, "[constructor]some-resource"])
hello_func = instance.get_func([IFACE, "[method]some-resource.hello"])

abort "exports not found in #{IFACE}" if [some_resource_constructor, hello_func].any?(&:nil?)

some_resource = some_resource_constructor.call
at_exit { some_resource.resource_drop }

hello_func(some_resource)
```

i made this after #619 

_store became store because the store parameter went from unused to used: resource lift/lower in convert.rs now needs it to wrap returned Val::Resource values with their owning store and to reject resources passed into calls on a different store.

### DIFF
--- a/ext/src/ruby_api/component/convert.rs
+++ b/ext/src/ruby_api/component/convert.rs
@@ -1,11 +1,12 @@
-use crate::ruby_api::component::component_namespace;
+use crate::ruby_api::component::{component_namespace, Resource};
 use crate::ruby_api::errors::ExceptionMessage;
 use crate::ruby_api::store::StoreContextValue;
 use crate::{define_rb_intern, err, error, not_implemented};
 use magnus::rb_sys::AsRawValue;
 use magnus::value::{IntoId, Lazy, ReprValue};
 use magnus::{
-    prelude::*, try_convert, value, Error, IntoValue, RArray, RClass, RHash, RString, Ruby, Value,
+    prelude::*, try_convert, value, Error, IntoValue, RArray, RClass, RHash, RString, Ruby,
+    TryConvert, Value,
 };
 use wasmtime::component::{Type, Val};
 
@@ -25,7 +26,7 @@ define_rb_intern!(
 pub(crate) fn component_val_to_rb(
     ruby: &Ruby,
     val: Val,
-    _store: Option<&StoreContextValue>,
+    store: Option<&StoreContextValue>,
 ) -> Result<Value, Error> {
     match val {
         Val::Bool(bool) => Ok(bool.into_value_with(ruby)),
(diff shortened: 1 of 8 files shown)


ext/src/ruby_api/component/resource.rs
```diff
@@ -0,0 +1,146 @@
...
+/// is no automatic finalizer: Ruby's GC may run at times where it would be
+/// unsafe to re-enter the {Store} (e.g. to invoke a guest-defined
+/// destructor), so failing to call {#resource_drop} will leak store-side
+/// resource-table state for the life of the {Store}.
+///
+/// Passing an +own<T>+ {Resource} into a function parameter transfers
+/// ownership of the underlying resource to the callee. Using the same
+/// {Resource} object afterwards (including calling {#resource_drop} on it)
+/// may then raise, since the store-side handle it pointed to no longer
+/// belongs to the host.
```
saulecabrera:
```suggestion
  /// Passing an +own<T>+ {Resource} into a function transfers ownership of the
  /// underlying resource to the callee. Once the call returns, this {Resource}
  /// object is a dangling handle; it still exists, but the store-side resource
  /// it referred to now belongs to the guest. Calling {#resource_drop} on it, or
  /// passing it into another call, may then raise. (+borrow<T>+ parameters do not
  /// transfer ownership, the same {Resource} stays valid and reusable.)

```

saulecabrera: Can we consider tightening the documentation here? Specifically to be more clear around the _when_ is the resource no longer valid: I find the usage of "afterwards" to be ambiguous in this context.

saulecabrera: Also, from what I can see there is no test covering this case? Can you add a test that covers it?

ext/src/ruby_api/component/convert.rs
```diff
@@ -284,8 +293,25 @@ pub(crate) fn rb_to_component_val(
...
+            match store.and_then(StoreContextValue::as_store) {
+                Some(current_store)
+                    if resource.store().as_value().as_raw()
+                        == current_store.as_value().as_raw() =>
+                {
+                    Ok(Val::Resource(resource.get()?))
+                }
+                Some(_) => {
+                    err!("Resource belongs to a different Store than the one being called into")
+                }
```
saulecabrera: Can we fold in the error case (when the stores differ) in the first branch?. This is minor, but I think it shows the error case clearer.

saulecabrera (commented):
Looks reasonable to me. Left some comments inline.

saulecabrera:
LGTM. We can land this after solving the conflicts.

teleyos:
> LGTM. We can land this after solving the conflicts.

I think the conflict is stale, I rebased on main to be able to fast-forward for a clean tree and it merges without conflict locally. The main side of the diff references context_mut()? (the pre-Magnus-migration API), which no longer exists on main - it was removed in #531.

I've never have this case so I don't know what to do, is it safe to accept my changes even if the conflict is stale or is it undefined behaviour ?

PR #624, opened 2026-07-05
Add Store#close

omohokcoj:
This PR adds `Store#close`, making it possible to free up Wasm memory deterministically.

```ruby
store = Wasmtime::Store.new(engine)
instance = Wasmtime::Instance.new(store, mod)
instance.invoke('process', input)

store.closed?  # => false
store.close    # => true  (frees Wasm memory, false if already closed)
store.closed?  # => true

# The store and anything created from it can no longer be used:
instance.invoke('process', input) # => raises Wasmtime::Error: Wasmtime::Store is closed
```


saulecabrera (approved):
LGTM, thanks!

